PR #9, opened 2013-07-18
Trim the contact address before adding it

Dinduks:
Because:
* Technically: addresses with spaces don't exist
* Functionally: a user may add an address with a space in it and wait for an
authorization that would never come


ibauersachs:

Dinduks:
Hey Ingo,
I signed the CLA. ;)

PR #13, opened 2013-12-12
Update dialog to add contact

tomka:
Based on a discussion on the mailing list I did some changes to the dialog used to add new contacts. I re-ordered the controls, renamed the labels, added tool tips. Additionally, I added faint gray prompts to the input fields (which will vanish if text is inserted) to guide the user. Below you will see a screen-shot of the changed dialog.

jitsi-new-add-account-dialog

**Edit:** After incorporating the last changes, the dialog now looks like this:

jitsi-new-add-account-dialog


ibauersachs:
Closing this in favor of #15

PR #15, opened 2014-01-17
Improve add contact dialog

tomka:
This pull request contains some commits discussed in pull request #13. It contains only those changes to the "Add contact" dialog were most people could agree on. Three changes were made: 1. Group field has been moved to bottom 2. Tool tips have been added and 3. example prompts have been added to the input fields. This is how the dialog now looks like:

jitsi_improved_add_contact_dialog

All other changes from #13 remain there for discussion.


ibauersachs:

tomka:
Thanks for getting back to this pull request! In case there is still interest in merging this: I signed your CLA back in 2014. Let me know if I should send it again.

ibauersachs:
@tomka there is a new CLA in effect since BlueJimp was acquired by Atlassian in spring '15. Everyone has/had to sign the new one (see links above and sorry for the confusion).

tomka:
Oh I see, thanks for the clarification. I will send a signed copy of the new CLA then.

PR #35, opened 2014-06-09
MainToolBar.call : Display chat transport resource name along with name

sandy-8925:
In Jitsi desktop version, in the case of Jabber/XMPP accounts (specifically Google Talk), if a contact is signed in through multiple clients (Jitsi, Google Hangouts), each client has a unique resource ID. (For Jitsi, jitsi-xxxxx and for Hangouts, Messaging-xxxxx)

In a chat panel, when the audio/video call button is clicked, a dropdown menu is shown, with the list of different XMPP addresses - however only the main address is shown (user@servername - e.g sandy.8925@gmail.com) - the resource portion (jitsi-xxxx,Messaging-xxxx) is not shown. (see attached picture)
jitsi_xmpp_dropdown_problem

So, when attempting to start an audio/video call through Jitsi, it is difficult to identify which endpoint/client needs to be selected since the resource is not shown - instead a list of multiple copies of the same email address is shown. This is important since Google Hangouts will not support audio/video calls with Jitsi, and the Jitsi client needs to be selected specifically.

### DIFF
--- a/src/net/java/sip/communicator/impl/gui/main/chat/toolBars/MainToolBar.java
+++ b/src/net/java/sip/communicator/impl/gui/main/chat/toolBars/MainToolBar.java
@@ -698,7 +698,7 @@ public class MainToolBar
             m.put(opSetClass, ct.getProtocolProvider());
 
             UIContactDetailImpl d = new UIContactDetailImpl(
-                                                ct.getName(),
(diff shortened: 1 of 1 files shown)


ibauersachs:
Closing in favor of #37

PR #37, opened 2014-06-20
VoiceCall Target resource

ArcticSnowSky:
This small changes allows you to select the resource you want to call (if multiples exist).
Before, you saw the target name (without resource) the amount of times of resources +1.
No matter which one chosen, called always the preferred one.

The hack in my solution is not the best one, but it works at least for xmpp voice.

I don't know if this problem would exist with other protocols too.

(more information on "MainToolBar.call : Display chat transport resource name along with name")


ibauersachs:

ArcticSnowSky:
iCLA signed May 29. 2016

PR #40, opened 2014-07-09
Better marketing

dandv:


ibauersachs:
Added some links and a changed description

PR #41, opened 2014-07-21
Could cause unnecessary exception

SkaceKamen:
Windows can have interface without MAC address, for example when using VPN.


ibauersachs:
Please sign the [CLA](https://jitsi.org/icla) if you haven't already done so.

SkaceKamen:
Signed

PR #66, opened 2015-01-27
Enable GTKLookAndFeel on Linux/FreeBSD globally

frikisama:
Right now, the native GTKLookAndFeel is only used when on a KDE session, which curiously enough is actually one of the few non-GTK-based DEs. The proposed changes enables GTKLookAndFeel to be used globally on Linux and FreeBSD, if the style is available, and gives users a much higher integration regardless their desktop environment of choice.

If there is a compelling reason on why not to enable it for other environments please tell me so I can look into another possible solution, because the out-of-the-box Jitsi UI experience on Linux could be much smoother. Thank you!

### DIFF
--- a/src/net/java/sip/communicator/impl/gui/UIServiceImpl.java
+++ b/src/net/java/sip/communicator/impl/gui/UIServiceImpl.java
@@ -897,19 +897,13 @@ public class UIServiceImpl
             {
                 try
                 {
-                    String kdeFullSession = System.getenv("KDE_FULL_SESSION");
-
-                    if ((kdeFullSession != null)
-                            && (kdeFullSession.length() != 0))
+                    for (LookAndFeelInfo lafi
+                            : UIManager.getInstalledLookAndFeels())
                     {
-                        for (LookAndFeelInfo lafi
-                                : UIManager.getInstalledLookAndFeels())
+                        if (gtkLookAndFeel.equals(lafi.getClassName()))
                         {
-                            if (gtkLookAndFeel.equals(lafi.getClassName()))
-                            {
-                                laf = gtkLookAndFeel;
-                                break;
-                            }
+                            laf = gtkLookAndFeel;
+                            break;
                         }
                     }
                 }


lyubomir:
Well... the matter at hand is complicated.

Firstly, the GTK look & feel is used either when it's the default (e.g. GNOME) or it integrates better with the DE (e.g. KDE). KDE has a (default) GTK theme which draws in a way similar to QT.

Secondly though, it turns out that the GTK look & feel in Java 7 & 8 uses GTK+ library in a way which causes recent versions of the latter to crash the process. So I have a change in my local source tree which switches away from the GTK look & feel to the OS-independent (Metal?) on Java 7 and/or 8.

The bottom line is that (1) I'm very reluctant to switch to the GTK look & feel on all DEs and (2) I'm thinking of pushing my switch away from it.

frikisama:
I'm not able to check right now, but last time I did even specifying the GTKLookAndFeel directly as an environment variable (export _JAVA_OPTIONS='-Dswing.defaultlaf=com.sun.java.swing.plaf.gtk.GTKLookAndFeel') when stating jitsi didn't work, and I had to actually fake a KDE session with "KDE_FULL_SESSION=bla" under XFCE to be able to change into the style. Maybe this is/was a bug?

Still, once I got it running with the GTK style, it ran just fine on Java 7 (OpenJDK/JRE) in Arch Linux. I'll test some more this week (Java 8, for instance).

frikisama:
Ok, just tested.

No way to set GTKLookAndFeel without setting KDE_FULL_SESSION. Without it, if the style is set in _JAVA_OPTIONS, it is detected and ignored. This should not be the case, all DEs should behave the same when specifically setting the style.

Also, with GTKLookAndFeel: Latest version with latest jre 7 (openjdk) works fine, except for a few places where color settings are ignored.

With Java 7

Same version (no recompile, don't know how much it affects) with latest jre 8 (openjdk) crashes, as you mentioned.

LeonardKoenig:
I'm actually opposing to hardcode GTK/Qt, as eg. KDE/Plasma 5 has *no* GTK-Theme yet. I'ld rather go for an options menu for selecting the theme. E.g. when I am on i3 (which uses neither, GTK nor Qt) I my want to use Qt (if this is possible). Same for LXQt and KDE, but when I'm on a GTK-desktop I'ld rather use GTK of course.

frikisama:
Removing it completely is, sure enough, another option (not my preferred one, but still...) to achieve consistency among DEs, because this was the actual idea behind this pull request. Of course, the correct way of handling this is for jitsi to honor the java options and settings from the system, and on top of that an even better solution is a style selection menu, as you suggested.

I won't have my PC available for some time (RMA...), so I won't be able to even test in a decent time frame because the only thing available to me right now is a sluggish netbook. Sorry!

MauroMombelli:
if system tary fails to load, but jitsi try anyway to use it; so if you close it to system tray (default close action for me) there is no way to close it but using pidof or similar.

I think you may prevent the usage of system tray if there is some error with it, so desktop without systemtray or similar does not have pitfall; for example if i enable "start iconized to system tray" there is no way to use the program unless i delete the configuration file. This should be fix regardless of the DE or solution used there

ibauersachs:
@MauroMombelli can you please open a separate issue for that? The tray has nothing to do with the look and feel (issues).

MauroMombelli:
Uff they just closed it and told me to come here -.-"

ibauersachs:
@MauroMombelli they is I and #120 is about the JVM crash on Plasma, which is related to the GTK look and feel mentioned here. The systray has nothing to do with it.

MauroMombelli:
#120 has been closed in favour of this. Anyway I'll try to open a new bug
as this is more general fallback

PR #70, opened 2015-03-14
Add SDES AES 192 and 256

tinours:
Add SDES AES 192 and 256


ibauersachs:
This will require an update of libjitsi too. I'll take a look when we did the 2.8 release. Is there any particular reason you need those ciphers? They're quite uncommon.

tinours:
Yeah new product to test i needed a sip client that support these two
ciphers. I manage to make it work in netbean but I had some problems with
the installer.

PR #108, opened 2015-04-06
Mention Google Hangouts in the description for SEO

dandv:
The goal is to increase discoverability for those who search GitHub for, say, [Google Hangouts encryption].

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 Jitsi
 =====
 
-Jitsi is a free open-source audio/video and chat communicator that supports protocols such as SIP, XMPP/Jabber, AIM/ICQ, ICQ, Yahoo! and many other useful features.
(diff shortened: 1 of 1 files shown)


bgrozev:
I don't think the particular phrasing is accurate. Jitsi is not "compatible" with hangouts in the sense that it can be used to connect to a hangout, and I don't understand how XMPP/Jabber could be considered "compatible" either.

dandv:
I'm not an XMPP expert, so feel free to change the phrasing. What I think would be useful is to allow users searching for secure chat software that works with their Google account to find Jitsi.

ibauersachs:
If we should mention this at all, I think it should be in Jitsi Meet as that is what comes closest to Hangouts.

dandv:
How about just Google Chat? As in, Jitsi can be used to communicate with someone who's using Google Chat. I'm just talking from the POV of a layman here, who has no idea what XMPP or Jabber are, but who knows about Google Chat/Talk (the IM-only capabilities).

fippo:
the layman who read the theverge article that google dropped xmpp with hangouts (the chat client, not the videochat)

PR #161, opened 2015-10-10
Fixes the contacts 'call button' disappearing on multiresource contacts

### LINKED ISSUE #111: Call button Dissappears if logged in trough Bitlbee
If an xmpp account is logged in trough Bitlbee and Jitsi at the same time, friends are not able to perform a call as the call button is greyed out. Both local and friend client are running version 2.8 on GNU/Linux.

459below:
This changes the behavior from "last connected resource" or "highest
priority resource" to "most capable resource" regarding the display of
buttons next to this contact. We now check every resource the contact
has connected and we will consider the one with the most capabilities.
This eliminates the very frustrating scenario in which the call
capabilities are disabled on a contact who connected additionally a
non-jingle client, thus resolves #111


459below:
I'm going to send in the contributor agreement in the next few days.

ibauersachs:
Thanks!
How does this behave now for two equally capable resources (e.g. chat)
- with the same priority
- with different priorities
- with different online statuses?

I'm a bit concerned about what happens when e.g. a mobile device resource is online and only chat capable and a desktop resource is call-capable, but away. The contact list shows the highest available online status, thus when you call a contact in this scenario it would ring on the desktop while said callee is potentially not even near his machine.

459below:
My intention was to reliably show the call buttons as long as there is such a client connected. This is done by comparing the size of the determined set of capabilities of the resources. The largest one wins. Jitsi/Jitsi-android has 35, while Psi, Adium and ChatSecure have 31. If there are two sets with the exact same size, the first one would win. My assumption here is, that it's most likely the same set. So it shouldn't be affected at all by priorities or statuses, only by resources. The call connects to the highest prioritized call capable resource, even if there is a non-call-capable resource with a higher priority. So the case you described with a lower prioritized call capable resource set on 'away' being 'covered' by a higher prioritized resource set on 'online' is how it would behave. However Jitsi displays the callee's resource which is being called to the caller. And in the extended tool-tips the priority and online status of all resources is listed, so the caller could see the away status of the resource they just called. Then they could send a message which would then be delivered to the callee's mobile device. That should be alright, shouldn't it?

Regarding same priorities and online statuses I noticed a different issue a few days back. I've already posted to jitsi-users [here] (http://lists.jitsi.org/pipermail/users/2015-October/010175.html) (2nd issue). I just noticed this seems to occur only if the two resources use the same priority.

While I was testing your proposed scenario I stumbled upon a certain constellation of resources and priorities in which it doesn't show the buttons, but the mobile indicator. I will look into that as well as in the other issue, but in a different thread.

ibauersachs:
I can't test and further comment for a couple of days. Please ping me if I don't get back to you by the end of next week. In the meantime, do you think you could make the behavior change optional? I'm not sure yet what would be the default then, but at least it would be configurable.

459below:
I can make this optional. I'll introduce a property for this, or did you mean an option in the GUI? I would like to point out, that Pidgin users [cannot](https://developer.pidgin.im/wiki/Protocol%20Specific%20Questions#HowcanIconfigureresourcepriority) specify their priorities. Also that Pidgin for Linux implements Jingle.

At the moment I'm still getting some weirdness by Psi on Linux, which I want to iron out.

ibauersachs:
A GUI option would be ideal. I'm thinking of a checkbox labeled something like "Use all resources when showing a contacts capabilities" in a new advanced configuration section for XMPP (similar to the one for SIP).

459below:
I set it to disabled as default. When it is disabled it uses the untouched code and otherwise it uses the new implementation. How should I proceed regarding translations? At the moment it's only residing in the generic resources.properties.

I'll commit this, after I've doubled checked the code and looked into that Psi issue some more.

screen shot 2015-10-26 at 14 01 11

ibauersachs:
We use [Pootle](http://translate.jitsi.org) for translations. You don't need to touch the other resource files (they'd get overwritten).

459below:
I doubled checked the code and tested it thoroughly. However the issue regarding the offline presence on same priority resources caused that weirdness I wanted to iron out. That line [here](https://github.com/459below/jitsi/blob/fix-callbutton/src/net/java/sip/communicator/impl/protocol/jabber/OperationSetContactCapabilitiesJabberImpl.java#L538) prevents that all caps nodes get removed if a rogue offline presence update comes in. At first I thought of it as a workaround, however after some thought it looks proper to me. 

I'm going to look into that presence issue now.

The CLA has been submitted on 17.10.

459below:
Seems that presence bug lies [here](https://github.com/459below/jitsi/blob/fix-callbutton/src/net/java/sip/communicator/impl/protocol/jabber/OperationSetPersistentPresenceJabberImpl.java#L1527). In case this ends up being exactly 0 again TreeSet will deem them equal and discard one of them.

I will verify this and if true put it into a different pull request.

src/net/java/sip/communicator/impl/protocol/jabber/OperationSetContactCapabilitiesJabberImpl.java
```diff
@@ -405,45 +444,80 @@ public void userCapsNodeAdded(String user, String node, boolean online)
...
-            String jid = StringUtils.parseBareAddress(user);
-            Contact contact = opsetPresence.findContactByID(jid);
-
-            // If the contact isn't null and is online we try to discover the
-            // new set of operation sets and to notify interested parties.
-            // Otherwise we ignore the event.
-            if (contact != null)
+        = parentProvider.getOperationSet(OperationSetPresence.class);
+        if (opsetPresence != null) {
+            if(LibJitsi.getConfigurationService()
```
ibauersachs: You cannot access the config service like this. It needs to be obtained from GeneralConfigPluginActivator.getConfigurationService()

src/net/java/sip/communicator/plugin/generalconfig/XMPPConfigForm.java
```diff
@@ -0,0 +1,99 @@
...
+import javax.swing.*;
+
+import org.jitsi.service.configuration.*;
+
+import net.java.sip.communicator.plugin.desktoputil.*;
+
+/**
+ * Implementation of the configuration form.
+ *
+ * @author Damian Minkov
```
ibauersachs: That would be you, not Damian :)

src/net/java/sip/communicator/impl/protocol/jabber/OperationSetContactCapabilitiesJabberImpl.java
```diff
@@ -405,45 +444,80 @@ public void userCapsNodeAdded(String user, String node, boolean online)
...
+                .getBoolean(
+                    XMPPConfigForm.PROP_XMPP_USE_ALL_RESOURCES_FOR_CAPABILITIES,
+                    XMPPConfigForm.USE_ALL_RESOURCES_FOR_CAPABILITIES_DEFAULT))
```
ibauersachs: The XMPPConfigForm is in a different bundle and may not be referenced in the protocol-jabber bundle. (It would e.g. break the Videobridge which doesn't have the generalconfig bundle). This probably only works because the constants are expanded at compile time. And because by convention you cannot reference impl-bundles in the UI, I currently only see the option of duplicating those constants in the protocol and the UI bundle.

src/net/java/sip/communicator/impl/protocol/jabber/OperationSetContactCapabilitiesJabberImpl.java
```diff
@@ -469,31 +544,54 @@ public void contactPresenceStatusChanged(
...
      */
-    public void fireContactCapabilitiesChanged(String user)
+    public void fireContactCapabilitiesChanged(String user, ArrayList<String> fullJids)
     {
-        OperationSetPresence opsetPresence
+        if(!LibJitsi.getConfigurationService()
+            .getBoolean(
+                XMPPConfigForm.PROP_XMPP_USE_ALL_RESOURCES_FOR_CAPABILITIES,
+                XMPPConfigForm.USE_ALL_RESOURCES_FOR_CAPABILITIES_DEFAULT)
+            ||fullJids.isEmpty())
```
ibauersachs: Same as above.

ibauersachs:
Okay, thanks, so:
- I need to verify the CLA with Emil as I don't have access to Docusign myself.
- Please address the inline comments.
- Please check your changes for violations of our [code convention](https://jitsi.org/Documentation/CodeConvention). There are some lines that are too long, braces on the same line, some missing spaces between operators. (But please don't reformat existing code, so far this PR is exemplar for that!)
- Enable your behavior by default. I want to get this into the nightlies to gather feedback.
- When you're done, please rebase the PR on master and squash the first three commits into one.
- Looking forward to your other PR, it probably shouldn't be a `TreeSet<>` but rather a `SortedList<>`. We had some issues with that before when the meaning of `Set` wasn't really understood.

src/net/java/sip/communicator/impl/protocol/jabber/OperationSetContactCapabilitiesJabberImpl.java
459below: Alright

src/net/java/sip/communicator/plugin/generalconfig/XMPPConfigForm.java
459below: Yes of course. That slipped through after I duplicated that class. My apologies to Damian.

459below:
I'll do all of that.

Regarding the squashing. I haven't done much collaborative work with git as of yet. The purpose of squashing the first three commits is to clean up the commit history since the second and third commit are tiny or should it even all be squashed into one big commit?

ibauersachs:
The first two commits (151e552f4ee8539f0ac0ba88f1e4a4391f871c44 and 13da0947f8a1a8e7ab9918741c051f51eaa2f7ff) logically belong together, it makes no sense to keep them as separate history on how the code became how it is. The third commit (b121edc96f6db8bf7014058cb8e199317ad1fde6) is a merge. Since this branch is not something other will work on, there's no need to not use rebase.
Same goes for the last commit (ad56640087885f0258a7bb0fad153b8194cc6ab6), you can get rid of that with rebasing (instead of merging).

src/net/java/sip/communicator/impl/protocol/jabber/OperationSetContactCapabilitiesJabberImpl.java
459below: I read up on the bundle convention in the Jitsi projects. So if I understand correctly, references like this need to be isolated to the respective bundle the class itself resides in. In this particular case I'd simply duplicate the constant to *OperationSetContactCapabilitiesJabberImpl*, since there isn't a more appropriate place for this.

ibauersachs: Yes.

src/net/java/sip/communicator/impl/protocol/jabber/OperationSetContactCapabilitiesJabberImpl.java
459below: I tried obtaining it via *GeneralConfigPluginActivator.getConfigurationService()*. That didn't work. *org.jitsi.service.libjitsi* is listed in *jabber.provider.manifest.mf*. Shouldn't this already comply with the bundle convention?

459below:
I've covered every point on the list. However that one inline comment was a tad problematic.

src/net/java/sip/communicator/impl/protocol/jabber/OperationSetContactCapabilitiesJabberImpl.java
ibauersachs: I might have confused the files. Here it's the JabberActivator (or JabberProtocolActivator or similar, I'm on my mobile).
It just always from the activator of the bundle the file lives in.

459below: Alright. Thanks. I'll try to get this done right away.

459below: OK, that worked.

459below:
OK. Now everything should be up to code.

src/net/java/sip/communicator/impl/protocol/jabber/OperationSetContactCapabilitiesJabberImpl.java
```diff
@@ -276,6 +289,41 @@ public OperationSetContactCapabilitiesJabberImpl(
...
+     * Each supported <tt>OperationSet</tt> capability is represented by a
+     * <tt>Map.Entry</tt> with key equal to the <tt>OperationSet</tt> class
+     * name and value equal to the respective <tt>OperationSet</tt> instance
+     */
+    protected Map<String, OperationSet> getLargestSupportedOperationSet(
+            ArrayList<String> fullJids)
+    {
+        Map<String, OperationSet> supportedOperationSets =
+            new HashMap<String, OperationSet>();
+        if (fullJids!=null){
```
ibauersachs: New line before { please

src/net/java/sip/communicator/impl/protocol/jabber/OperationSetContactCapabilitiesJabberImpl.java
```diff
@@ -276,6 +289,41 @@ public OperationSetContactCapabilitiesJabberImpl(
...
+            for (String fullJid : fullJids)
+            {
+                Map<String, OperationSet> newSupportedOperationSets=
+                getSupportedOperationSets(fullJid, true);
+                if (newSupportedOperationSets.size()>
+                supportedOperationSets.size())
```
ibauersachs: Please use an indent here, it's hard to read like this

src/net/java/sip/communicator/impl/protocol/jabber/OperationSetContactCapabilitiesJabberImpl.java
```diff
@@ -405,45 +455,83 @@ public void userCapsNodeAdded(String user, String node, boolean online)
...
                         contact,
-                        ContactCapabilitiesEvent.SUPPORTED_OPERATION_SETS_CHANGED,
-                        getSupportedOperationSets(user,
-                            online));
+                        ContactCapabilitiesEvent.
+                        SUPPORTED_OPERATION_SETS_CHANGED,
+                        getLargestSupportedOperationSet(fullJids));
                 }
-                else
+            } else {
```
ibauersachs: New lines after } and before {

src/net/java/sip/communicator/impl/protocol/jabber/OperationSetContactCapabilitiesJabberImpl.java
```diff
@@ -469,31 +558,57 @@ public void contactPresenceStatusChanged(
...
+                if(contact != null)
+                {
+                    fireContactCapabilitiesEvent(
+                        contact,
+                        ContactCapabilitiesEvent.
+                        SUPPORTED_OPERATION_SETS_CHANGED,
+                        getSupportedOperationSets(user, online));
+                }
+            }
+        } else {
```
ibauersachs: Same

459below:
Sorry for missing those. This coding style is very different from my own. The next one should be better :)

ibauersachs:
Merged. Thank you!
Do you want to make an announcement on dev so that people know the difference?

459below:
Unfortunately I've missed one test case. This introduced a bug. The offline bug threw me off track here and I falsely blamed it for this (an unhandled exception looks very similar). If a contact **actually** goes offline, we do not grant him the offline capabilities, which among others is *sending offline messages*. We can't open the chat window, however we could write, if the window is already open.

I fixed it in this commit https://github.com/459below/jitsi/commit/f393b01befb08b2e974f4e741b3b01d5a30051cb
Annoyingly enough I actually did take it into account [here] (https://github.com/459below/jitsi/blob/f393b01befb08b2e974f4e741b3b01d5a30051cb/src/net/java/sip/communicator/impl/protocol/jabber/OperationSetContactCapabilitiesJabberImpl.java#L574), but not [here] (https://github.com/459below/jitsi/blob/f393b01befb08b2e974f4e741b3b01d5a30051cb/src/net/java/sip/communicator/impl/protocol/jabber/OperationSetContactCapabilitiesJabberImpl.java#L493)

ibauersachs:
Can you please create a separate PR for that commit?
And please uses a space around operators (and assignments in the future too).

PR #164, opened 2015-10-13
Fix condition precedence

treble-snake:
Now `else` is related to inner `if`, which is obviously a bug. This way `srcContact ` will always be `null`.


ibauersachs:
Could you please sign our [contributor agreement](https://jitsi.org/icla)? Thanks!

treble-snake:
@ibauersachs sure thing, but couldn't you tell - should I use real name or github account name there?

ibauersachs:
Please use your real name. If we're unable to associate that with your Github name, please send a private e-mail to me.

treble-snake:
@ibauersachs Done! You can associate the agreement with my GitHub name by E-mail :)

PR #171, opened 2015-10-16
Implements arbitrary packet logging configuration.

gpolitis:
This PR complements https://github.com/jitsi/libjitsi/pull/64

### DIFF
--- a/m2/jitsi-packetlogging/pom.xml
+++ b/m2/jitsi-packetlogging/pom.xml
@@ -23,6 +23,10 @@
       <artifactId>libjitsi</artifactId>
     </dependency>
     <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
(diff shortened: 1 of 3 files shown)


ibauersachs:
Could you please try doing a sip call before you merge this? I'm a bit afraid that the bundle-plugin somehow sometimes doesn't catch all necessary imports from fmj.

gpolitis:
I will, thanks for thinking about this Ingo. I'm trying to understand why you're referring to fmj, since there's no dependency between fmj and the packet logging service implementation. Could you please elaborate a little bit further on this?

ibauersachs:
Oh, sorry, that was about updating the libjitsi binary. The last time I updated it, I think I got an outdated fmj-snapshot from our github-mvn-repo and thus maven-bundle-plugin didn't catch all necessary OSGi imports that needed to be put into the libjitsi manifest.

PR #183, opened 2015-11-10
Introduce SmackInteroperabilityLayer for abstracting different Smack implementations (3 and 4)

mksh:
- base class for abstracting Smack
- implementation for SmackV3

### DIFF
--- a/src/net/java/sip/communicator/impl/protocol/jabber/ProtocolProviderFactoryJabberImpl.java
+++ b/src/net/java/sip/communicator/impl/protocol/jabber/ProtocolProviderFactoryJabberImpl.java
@@ -21,6 +21,7 @@ import java.util.*;
 
 import net.java.sip.communicator.service.protocol.*;
 
+import net.java.sip.communicator.service.protocol.jabber.*;
 import org.jivesoftware.smack.provider.*;
 import org.jivesoftware.smack.util.*;
 import org.osgi.framework.*;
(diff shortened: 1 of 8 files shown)


ibauersachs:
Would someone mind explaining to me what this is good for?

bgrozev:
Ingo, we are adding BOSH support to jitsi-hammer, which uses the smack extensions from the Jitsi bundle. For this we updated jitsi-hammer's smack to version 4, and @mksh created an abstraction layer, so we can reuse the code and facilitate the potential future migration for Jitsi. A smack4 implementation of the layer will be added to jitsi-hammer soon.

ibauersachs:
Thanks Boris. I never really looked into it and thus cannot estimate the amount of work required. But wouldn't it make more sense to upgrade Jitsi to Smack V4 for the benefit of all involved projects instead of writing abstraction layers and procrastinating the inevitable?

bgrozev:
I can't estimate the effort accurately either, but I would tend to think it would be substantial. I also suspect it would introduce bugs, which we wouldn't have the time to fix. In any case, I don't think adding this layer is procrastination -- rather, it is an actual step towards migrating, since part of the code is now v4-ready (or at least v4.0.7-ready).

damencho:
I just wanted to check which of our patches had been integrated there, I suspect most of them, but I can spent some time on this and report it.

PR #184, opened 2015-11-10
Fixes the contacts 'message button' disappearing on offline contacts

459below:
This is a fixup to https://github.com/jitsi/jitsi/pull/161 This introduced a bug in which a contact which just went offline to not have *messaging capabilities*. This is needed for opening the chat window to send offline messages. One clause was missing, causing the new behavior to check offline contacts for their capabilities. More specifically it checked their online resources and this will always be a null set on an offline contact. With this it now uses the old behavior for offline contacts, which returns a preset set of capabilities (e.g. messaging).


459below:
I'm aware that those two statements are actually identical, but negated. There is no intention behind this. Should I reorder this or leave it for now?

ibauersachs:
Thanks. The condition doesn't matter.

459below:
My pleasure! So you want me to send an email to the dev list in which I explain what this new option does?

PR #185, opened 2015-11-15
Fix TreeSet's Comparator in Jabber Presence class

### LINKED ISSUE #182: Contacts with several same prioritized resources sometimes show wrong overall status
I'm opening this separate issue thread for the bug mentioned [here] (http://lists.jitsi.org/pipermail/users/2015-October/010175.html) and [here] (https://github.com/jitsi/jitsi/pull/161)

The bug lies [here] (https://github.com/459below/jitsi/blob/fix-callbutton/src/net/java/sip/communicator/impl/protocol/jabber/OperationSetPersistentPresenceJabberImpl.java#L1527)

459below:
This fixes the Presence TreeSet's Comparator. If the comparator can't
determine a greater or less than relationship, it will sort by the name
of the resources lexicographically.
A (mathematical) set is a collection of distinct objects. A TreeSet's
compare function determines relation of two objects. It also determines
equality and since this is a Set, it would effectively replace the old
one in the set, if the function returns 0. This method wasn't fully
equals consistent, meaning different objects were deemed equal by the
comparator. That lead to an inconsistency in the representation of
presence objects and resources that contact has connected. The resources
not represented in this TreeSet would no longer be able to let the
contact appear as Online.

This resolves #182

### DIFF
--- a/src/net/java/sip/communicator/impl/protocol/jabber/OperationSetPersistentPresenceJabberImpl.java
+++ b/src/net/java/sip/communicator/impl/protocol/jabber/OperationSetPersistentPresenceJabberImpl.java
@@ -1528,6 +1528,19 @@ public class OperationSetPersistentPresenceJabberImpl
                                         o2, parentProvider).getStatus()
                                       - jabberStatusToPresenceStatus(
                                             o1, parentProvider).getStatus();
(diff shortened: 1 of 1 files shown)


ibauersachs:
Thanks!

PR #190, opened 2015-11-20
Program name prepending for log formatter

mksh:
This is aimed to use with Jitsi components to make them identify their names when shipping logs into, for instance, syslog.


ibauersachs:
Lgtm

PR #204, opened 2016-01-14
Directimage fix

tklein:
I had problems with some image URLs not being replaced correctly on Windows.

Our internal XMPP server uses URLs of the form
   https://im.example.org:12345/abc/def/ghi/img.jpg
which weren't replaced. Other URLs worked fine though.

Tried the fix on Linux and it works perfectly fine.


ibauersachs:
If you haven't already done so, could you please sign the [CLA](https://jitsi.org/icla)?

tklein:
Already did yesterday, in case it's somehow not submitted pls drop me a line.

PR #212, opened 2016-01-28
Support running run.sh from any current working directory

andrey-utkin:
When stable package installs to user's $HOME/jitsi, currently, run.sh starts successfully only when CWD is $HOME/jitsi.
Don't know for which other cases run.sh is used, please take care to check.

### DIFF
--- a/resources/install/linux/run.sh
+++ b/resources/install/linux/run.sh
@@ -1,4 +1,5 @@
 mkdir -p $HOME/.sip-communicator/log
+cd `dirname $0`
 
 export PATH=$PATH:native
 export JAVA_HOME=jre


ibauersachs:
I'm not sure if cd in a script is a good idea. @damencho ?

narfman0:
I think it's better to run everything within the script relatively, e.g.:
export SCRIPTPATH=$( cd $(dirname $0) ; pwd -P )
then every subsequent relatively pathed script invocation/variable can be prefixed with SCRIPTPATH, and you don't trash your cwd. no pushd or popd, either.

damencho:
@ibauersachs I agree with @narfman0, in jicofo and jvb we already have:
SCRIPT_DIR="$(dirname "$(readlink -f "$0")")"

ibauersachs:
@andrey-utkin could you please rework this to not use `cd`?

andrey-utkin:
Feel free to rework in any way you like without worries about my authorship.

ibauersachs:
I don't have time to work on this. Please reopen or create a new PR if you can do the changes as suggested in the comments.

PR #235, opened 2016-03-07
Pipe does not work in windows makefile with Mingw and Msys so we came up with another solution that does.  This also removes the last remaining dependency on msys (sed) as far as we can tell.

powerpbx:


powerpbx:
This is just our first try.  If anyone has a better way to do this we we are open to suggestions.

src/net/java/sip/communicator/impl/version/NightlyBuildID.java
```diff
@@ -27,5 +27,5 @@
     /**
      * The build ID.
      */
-    public static final String BUILD_ID="0.build.by.SVN";
+    public static final String BUILD_ID="1";
 }
```
damencho: When invoking ant you just need to pass -Dlabel="1" and this is going to be set as build version.

ibauersachs:
You should work on a branch, not on master. And for a pull request you *need* to do that.

resources/install/build.xml
```diff
@@ -573,7 +573,7 @@
                 todir="${light.dir}"
                 overwrite="true" />
 
-        <property name="wix.home" value="C:\Program Files\WiX" />
+        <property name="wix.home" value="C:\wix310" />
```
damencho: This also can be controlled with passing parameters to ant as -Dwix.home="C:\wix310".

powerpbx:
Thanks for the info.

PR #240, opened 2016-03-21
Use string comparison instead of int comparison

coypoop:
Currently if the architecture is defined differently (such
as 'amd64' used in NetBSD), this comparison fails with:
[: amd64: bad number


coypoop:
This appears elsewhere for Debian and Linux, it might be worth fixing those too, in case someone wants to build for architectures other than Intel ones.

ibauersachs:

coypoop:
Done

PR #277, opened 2016-06-23
code clean up

nikvaessen:
replaced if/else constructs with switches
replaced foreach loops
added log entries
fixed order in static fields
switched comparisons of Strings to not generate null pointers


ibauersachs:
Nik, please rebase your changes on top of master instead of merging them into your branch.

PR #281, opened 2016-07-06
Updates libjitsi and ice4j.

bgrozev:


ibauersachs:
What's the policy on updating the pinned versions in m2/pom.xml?

bgrozev:
I don't know about the policy, but I think it makes sense to update them. I'll do so in another PR, because I want to recreate this one anyway (to include the versions used in the commit message).

PR #312, opened 2017-01-17
Adds an option for setting audio rtp translator and avoids mixing audio.

damencho:


damencho:
This approach is not working, closing PR.

PR #380, opened 2017-07-18
fix incorrect SourcePacketExtension parse

bbaldino:
…tExtension (and was missing the rid attribute)

### DIFF
--- a/test/net/java/sip/communicator/impl/protocol/jabber/extensions/colibri/ColibriIQProviderTest.java
+++ b/test/net/java/sip/communicator/impl/protocol/jabber/extensions/colibri/ColibriIQProviderTest.java
@@ -1,0 +1,185 @@
+/*
+ * Jitsi, the OpenSource Java VoIP and Instant Messaging client.
+ *
+ * Copyright @ 2015 Atlassian Pty Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.java.sip.communicator.impl.protocol.jabber.extensions.colibri;
+
+import junit.framework.TestCase;
+import net.java.sip.communicator.impl.protocol.jabber.SmackV3InteroperabilityLayer;
+import net.java.sip.communicator.service.protocol.jabber.AbstractSmackInteroperabilityLayer;
+import org.jivesoftware.smack.packet.IQ;
+import org.xmlpull.v1.XmlPullParser;
+import org.xmlpull.v1.XmlPullParserException;
+import org.xmlpull.v1.XmlPullParserFactory;
+
+import java.io.IOException;
+import java.io.StringReader;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class ColibriIQProviderTest extends TestCase
+{
+    String testXml = "\n" +
+            "<iq type=\"set\" to=\"jvb-0.brian2.jitsi.net\" from=\"focus@auth.brian2.jitsi.net/focus358522582823562\" id=\"qV8NP-1407\">" +
+              "<conference xmlns=\"http://jitsi.org/protocol/colibri\" id=\"cce6f2fe74002273\" name=\"test\" gid=\"ff62ef\">" +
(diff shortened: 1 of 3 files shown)


test/net/java/sip/communicator/impl/protocol/jabber/extensions/colibri/ColibriIQProviderTest.java
```diff
@@ -0,0 +1,190 @@
...
+import org.xmlpull.v1.XmlPullParserFactory;
+
+import java.io.IOException;
+import java.io.StringReader;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class ColibriIQProviderTest extends TestCase
+{
+    String testXml = "\n" +
```
ibauersachs: Can't you place that in a resource file next to the test?

bbaldino: sure, is there an example of that somewhere i can follow?

ibauersachs: I don't think so, since we're lacking unit tests... You should be able to simply place it in a file next to the class, and then load it via getClass().getResourceAsStream("/net/java/sip/communicator/impl/protocol/jabber/extensions/colibri/testXml.xml")

bbaldino: oh that sounds cool...yeah i'll try that

bbaldino: @ibauersachs gave this a shot but unfortunately it looks like this doesn't play nicely with a properly-formatted xml resource file (i.e. with indentation).  From what I can tell it would require it to be one continuous block of text, which is a bit of a bummer.  XmlPullParser seems to choke on the newlines/spaces.  I'll dig around a bit and see if there's another way (because I do think this would be a nice separation).  Let me know if you have any ideas.

test/net/java/sip/communicator/impl/protocol/jabber/extensions/colibri/ColibriIQProviderTest.java
```diff
@@ -0,0 +1,190 @@
...
+    public void setUp()
+    {
+        try {
+            xmlPullParserFactory = XmlPullParserFactory.newInstance();
+            xmlPullParserFactory.setNamespaceAware(true);
+
+            xmlPullParser = xmlPullParserFactory.newPullParser();
+
+        } catch (XmlPullParserException e) {}
+        AbstractSmackInteroperabilityLayer.setImplementationClass(
```
ibauersachs: While we're at it, please move the braces to new lines (and the identation is wrong too).

test/net/java/sip/communicator/impl/protocol/jabber/extensions/colibri/ColibriIQProviderTest.java
ibauersachs: Since this is simply for testing, how about this:
1. parse the XML from the file using a full-blown parser
2. the write it out again without whitespace/newlines
3. let XmlPullParser read from the output of 2.

bbaldino: this is frustratingly lengthy as well...i used the javax parser but when writing it out it formats it neatly (with newlines and spaces), so to do step 2 would involve manually walking through the xml tree and concatenating each bit together.  what about for now moving the string (as it originally was here) to a separate java file, still defined as it is now?   i like the idea of the separate xml file, but not sure it'll end up being worth the extra code.

test/net/java/sip/communicator/impl/protocol/jabber/extensions/colibri/ColibriIQProviderTest.java
bbaldino: will fix this

test/net/java/sip/communicator/impl/protocol/jabber/extensions/colibri/ColibriIQProviderTest.java
ibauersachs: *facepalm*. Just another reason why I hate Java. Leave it as it is.

ibauersachs: C# ftw:
```
XDocument.Parse("<root>\r\n    <asdf/>    </root>").ToString(SaveOptions.DisableFormatting)
```

bbaldino:
i was about to add asserts in the catch blocks but i think it works better to just have the test method throw.

bbaldino:
@ibauersachs this look alright now?

ibauersachs:
Just the curly braces position in the setup() method.

bbaldino:
alright hopefully that is all of it.  please squash when you merge.

PR #410, opened 2017-10-11
Introduces stats-id xmpp extension and colibri changes around it

damencho:


src/net/java/sip/communicator/impl/protocol/jabber/extensions/colibri/ColibriConferenceIQ.java
```diff
@@ -2232,14 +2292,26 @@ public boolean removeSctpConnection(SctpConnection connection)
         private String id;
 
         /**
+         * The 'stats-id' of this <tt>Endpoint</tt>.
+         */
+        private String statsId;
+
+        /**
+         * The transport element of this <tt>ChannelBundle</tt>.
```
bgrozev: s/ChannelBundle/Endpoint/

src/net/java/sip/communicator/impl/protocol/jabber/ProtocolProviderServiceJabberImpl.java
```diff
@@ -1746,6 +1746,11 @@ protected void initialize(EntityBareJid screenname,
                 AvatarUrl.NAMESPACE,
                 new AvatarUrl.Provider());
 
+            ProviderManager.addExtensionProvider(
+                StatsId.ELEMENT_NAME,
```
ibauersachs: AvatarUrl is registered above, if you want StatsId parsing, create a distinct Provider class for it

src/net/java/sip/communicator/impl/protocol/jabber/extensions/colibri/ColibriBuilder.java
```diff
@@ -357,39 +355,41 @@ public boolean addBundleTransportUpdateReq(
...
-        // Note: if the request already contains a bundle with the same ID, the
-        // OLD one is kept.
-        boolean hasAnyChanges = request.addChannelBundle(bundleUpdate);
+        // Note: if the request already contains a endpoint with the same ID,
+        // the OLD one is kept.
+        boolean hasAnyChanges = request.addEndpoint(endpointUpdate);
         if (!hasAnyChanges)
         {
-            logger.warn("A channel-bundle update has been lost (an instance "
+            logger.warn("A endpoint update has been lost (an instance "
```
ibauersachs: An

src/net/java/sip/communicator/impl/protocol/jabber/extensions/colibri/ColibriConferenceIQ.java
```diff
@@ -2271,13 +2361,64 @@ public void setDisplayName(String displayName)
...
+         * Sets the stats ID of this <tt>Endpoint</tt>.
+         * @param statsId the stats ID to set.
+         */
+        public void setStatsId(String statsId)
+        {
+            this.statsId = statsId;
+        }
+
+        /**
+         * Appends the XML <tt>String</tt> representation of this
```
ibauersachs: If you reference other classes, please start using {@link ClassName} instead of <tt>, this creates navigatable links and is safe for refactoring as IDEs will recognize the reference.

src/net/java/sip/communicator/impl/protocol/jabber/extensions/colibri/ColibriConferenceIQ.java
```diff
@@ -2271,13 +2361,64 @@ public void setDisplayName(String displayName)
...
+            IQChildElementXmlStringBuilder xml)
+        {
+            xml.halfOpenElement(ELEMENT_NAME)
+                .attribute(ID_ATTR_NAME, id);
+
+            xml.optAttribute(DISPLAYNAME_ATTR_NAME, displayName);
+            xml.optAttribute(STATS_ID_ATTR_NAME, statsId);
+
+
+            IceUdpTransportPacketExtension transport = getTransport();
```
ibauersachs: You could probably just call addExtension(...) and let Smack deal with adding the extension element xml (and forget about the condition).

damencho: I don't get this ... I don't see addExtension ... as this is generating the xml ...?

src/net/java/sip/communicator/impl/protocol/jabber/extensions/StatsId.java
```diff
@@ -0,0 +1,121 @@
...
+        return NAMESPACE;
+    }
+
+    /*
+     * Returns xml representation of this extension.
+     * @return xml representation of this extension.
+     */
+    public String toXML()
+    {
+        final StringBuilder buf = new StringBuilder();
```
ibauersachs: If possible, use Smack's `XmlStringBuilder`

src/net/java/sip/communicator/impl/protocol/jabber/extensions/colibri/ColibriConferenceIQ.java
```diff
@@ -2271,13 +2361,63 @@ public void setDisplayName(String displayName)
         }
 
         /**
+         * Sets the transport element of this <tt>ChannelBundle</tt>.
+         * @param transport the transport to set.
+         */
+        public void setTransport(IceUdpTransportPacketExtension transport)
+        {
+            this.transport = transport;
```
ibauersachs: `addExtension(transport);` (+ null handling, calling twice, etc.)
Saving transport to a field is then optional, but more convenient than retrieving it via `getExtension(..., ...)`

src/net/java/sip/communicator/impl/protocol/jabber/extensions/colibri/ColibriConferenceIQ.java
```diff
@@ -2271,13 +2361,63 @@ public void setDisplayName(String displayName)
...
+        public IQChildElementXmlStringBuilder toXML(
+            IQChildElementXmlStringBuilder xml)
+        {
+            xml.halfOpenElement(ELEMENT_NAME)
+                .attribute(ID_ATTR_NAME, id);
+
+            xml.optAttribute(DISPLAYNAME_ATTR_NAME, displayName);
+            xml.optAttribute(STATS_ID_ATTR_NAME, statsId);
+
+            IceUdpTransportPacketExtension transport = getTransport();
```
ibauersachs: Forget about all that when `addExtension` is called from within `setTransport`, just call `setEmptyElement` here.

NEVER call `closeElement`, Smack does this for you. Have a look at [getChildElementXML()](https://github.com/igniterealtime/Smack/blob/master/smack-core/src/main/java/org/jivesoftware/smack/packet/IQ.java#L156), this might clear things up. It behaves quite differently than just doing plain XML construction.

src/net/java/sip/communicator/impl/protocol/jabber/extensions/colibri/ColibriConferenceIQ.java
ibauersachs: See new comments.

src/net/java/sip/communicator/impl/protocol/jabber/extensions/colibri/ColibriConferenceIQ.java
damencho: But this Endpoint is not an iq ... this is what confuses me... I was reusing what we already had there and I've added the stuff with the same logic that already existed ... Maybe you are talking about rewriting the whole ColibriConferenceIQ, which makes sense, but maybe we need to schedule this work and do it in another PR not as part of this one.

src/net/java/sip/communicator/impl/protocol/jabber/extensions/colibri/ColibriConferenceIQ.java
ibauersachs: A unit test based how the element should look like wouldn't hurt (similar to e.g. `MuteIqProviderTest` in Jicofo, which should actually be somewhere here in Jitsi)

damencho: Ok, so you want to refactor all the logic in the ColibriConferenceIQ before doing these changes first?

ibauersachs: No, why?

src/net/java/sip/communicator/impl/protocol/jabber/extensions/colibri/ColibriConferenceIQ.java
ibauersachs: I started looking at that on the phone and just saw the familiar getChildElementXml - and how that's built there wouldn't work if Endpoint would be an IQ, which it isn't. Sorry about the noise.

PR #412, opened 2017-10-13
Updates stats-id in endpoints.

damencho:


src/net/java/sip/communicator/impl/protocol/jabber/extensions/colibri/ColibriConferenceIQ.java
```diff
@@ -2724,4 +2669,30 @@ public void setPort(int port)
             this.port = port;
         }
     }
+
+    public static class TransportCommon
```
bgrozev: How about "TransportContainer"?

PR #423, opened 2017-11-07
Colibri fixes and cleanup

bgrozev:


ibauersachs:
This has been open for a while. Are you waiting for a review?

bgrozev:
I had forgotten about this one, but yes, I'm looking for a reviewer. @bbaldino , @damencho ?

src/net/java/sip/communicator/impl/protocol/jabber/extensions/colibri/ColibriBuilder.java
```diff
@@ -72,6 +72,132 @@
...
+    }
+
+    /**
+     * Copies the contents of a {@link RtpDescriptionPacketExtension} to a
+     * {@link ColibriConferenceIQ.Channel}.
+     * @param description the {@link RtpDescriptionPacketExtension} from which
+     * to copy.
+     * @param channel the {@link ColibriConferenceIQ.Channel} to which to copy.
+     */
+    private static boolean copyDescription(
```
bbaldino: would this be better as a clone method inside of RtpDescriptionPacketExtension?

bbaldino: after reading this some more, i'm wondering if 'copy' feels accurate...we don't actually make a copy of the description, more extract fields from it to put in the channel?  is this something more like `parseRtpDescriptionToChannel` or `addPayloadTypesAndHdrExtensions`?

src/net/java/sip/communicator/impl/protocol/jabber/extensions/colibri/ColibriBuilder.java
```diff
@@ -233,17 +364,29 @@ public boolean addAllocateChannelsReq(
...
+            {
+                requestChannel.setEndpoint(endpointId);
+            }
             requestChannel.setInitiator(peerIsInitiator);
 
-            if (useBundle)
+            if (useBundle && endpointId != null)
             {
-                requestChannel.setChannelBundleId(endpointName);
+                requestChannel.setChannelBundleId(endpointId);
```
bbaldino: no bundle id for octo channels because we're not using ice there, right?

src/net/java/sip/communicator/impl/protocol/jabber/extensions/colibri/ColibriBuilder.java
```diff
@@ -406,16 +528,14 @@ public boolean addBundleTransportUpdateReq(
...
-        }
+        request.addChannelBundle(channelBundleRequest);
 
-        return hasAnyChanges;
+        // Note that we don't actually check whether the addition of the bundle
+        // made any changes. We return true, because it is safe. It might lead
+        // us to send an unnecessary request, but it will not fail to send a
+        // request when we need it. We may need to add a check for modification
+        // of the content of the channel bundle extension.
+        return true;
```
bbaldino: was there a reason to change the logic here to this?

src/net/java/sip/communicator/impl/protocol/jabber/extensions/colibri/ColibriAnalyser.java
```diff
@@ -169,7 +172,7 @@ public static ColibriConferenceIQ getResponseContents(
 
                     bundleId = readChannelBundle(channelResponse, bundleId);
 
-                    endpoints.add(channelResponse.getEndpoint());
+                    endpointIds.add(channelResponse.getEndpoint());
```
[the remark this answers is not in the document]
bgrozev: Well, the XML attribute is called "endpoint" and I think it's too late to change it (also, I'm too lazy to also update jicofo and jitsi-videobridge).

src/net/java/sip/communicator/impl/protocol/jabber/extensions/colibri/ColibriBuilder.java
bgrozev: copyDescriptionToChannel?

src/net/java/sip/communicator/impl/protocol/jabber/extensions/colibri/ColibriBuilder.java
bgrozev: Correct

src/net/java/sip/communicator/impl/protocol/jabber/extensions/colibri/ColibriBuilder.java
bbaldino: my only hesitation with 'copy' was that we don't actually do a deep copy of the header extensions (just add the references), but if you don't think that's a big deal then i'm fine with `copyDescriptionToChannel`

src/net/java/sip/communicator/impl/protocol/jabber/extensions/colibri/ColibriAnalyser.java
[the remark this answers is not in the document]
bbaldino: k

src/net/java/sip/communicator/impl/protocol/jabber/extensions/colibri/ColibriBuilder.java
bgrozev: Yeah, because of the updated semantics of request.addChannelBundle(). It used to keep the old value, now it replaces the old value with the new value (which I think makes more sense).

So if we execute this twice, we can't easily check if the second call resulted in an update or not.

src/net/java/sip/communicator/impl/protocol/jabber/extensions/colibri/ColibriBuilder.java
```diff
@@ -308,61 +470,21 @@ public boolean addAllocateChannelsReq(
...
-            ColibriConferenceIQ.Channel channel)
-    {
-        RtpDescriptionPacketExtension description
-            = content.getFirstChildOfType(
-                    RtpDescriptionPacketExtension.class);
-        if (description != null)
+        else if (useBundle && endpointId == null)
         {
-            return copyDescription(description, channel);
+            logger.error("XXX");
```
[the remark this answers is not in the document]
bgrozev: Fixed.

src/net/java/sip/communicator/impl/protocol/jabber/extensions/colibri/ColibriBuilder.java
bbaldino: ok, so `addChannelBundle` will replace an existing bundle if there was one?  maybe we should rename it to `setChannelBundle` ?

bgrozev: What would you expect addChannelBundle to do in this case? Replacing is consistent with Map.put()

bbaldino: i'd expect `addChannelBundle` to add this new bundle, without removing anything that was existing. `setChannelBundle` to me says: "set this as THE one and only channel bundle, replacing anything that was there".  put i think matches 'set' more closely than add.

bgrozev: But there isn't only one channel bundle. In XML there's multiple child elements, and in java there's a map. I'd find a chain of "setChannelBundle" calls quite strange.

bbaldino: ok, when you said "It used to keep the old value, now it replaces the old value with the new value" it sounded like whenever 'add' got called, it replaced the old one with the current one, meaning there could only be one bundle set.  did i misunderstand that?

bbaldino: gotcha.

PR #431, opened 2017-12-21
Propfile config bug

networkinterfacecontroller:
I created a patch for a bug that was preventing the enabling of prop file configuration on Windows. The code to load from jitsi-defaults.properties and jitsi-default-overrides.properties was moved to JdbcConfigService.java, so the condition Windows systems needed to meet to enable prop config would always fail. This is tested and in production internally.

### DIFF
--- a/src/net/java/sip/communicator/impl/configuration/ConfigurationActivator.java
+++ b/src/net/java/sip/communicator/impl/configuration/ConfigurationActivator.java
@@ -21,7 +21,6 @@ import com.sun.jna.*;
 
 import net.java.sip.communicator.util.ServiceUtils;
 
-import org.jitsi.impl.configuration.ConfigurationServiceImpl;
 import org.jitsi.service.configuration.*;
 import org.jitsi.service.fileaccess.*;
 import org.jitsi.service.libjitsi.*;
--- a/src/net/java/sip/communicator/impl/configuration/ConfigurationActivator.java
+++ b/src/net/java/sip/communicator/impl/configuration/ConfigurationActivator.java
@@ -21,6 +21,7 @@ import com.sun.jna.*;
 
 import net.java.sip.communicator.util.ServiceUtils;
 
+import org.jitsi.impl.configuration.ConfigurationServiceImpl;
 import org.jitsi.service.configuration.*;
 import org.jitsi.service.fileaccess.*;
 import org.jitsi.service.libjitsi.*;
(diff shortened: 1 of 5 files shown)


src/net/java/sip/communicator/impl/configuration/ConfigurationActivator.java
```diff
@@ -89,19 +203,16 @@ public void start(BundleContext bundleContext)
 
     private boolean usePropFileConfigService(BundleContext bundleContext)
     {
-        if (Boolean.getBoolean(PNAME_USE_PROPFILE_CONFIG))
+        if (Boolean.valueOf(defaultProperties.get(PNAME_USE_PROPFILE_CONFIG)))
```
ibauersachs: No. This is intentionally a system property and not a Jitsi configuration property. It is only intended to be used by Jitsi Meet components (e.g. Jigasi) embedding some parts of Jitsi Desktop.

ibauersachs:
I don't mind the idea of a Jitsi configuration property in -defaults or -default-overrides that disables the database based store. But if at all, then
- the system property for disabling the database service needs to stay
- all you want in the activator is this specific property, no need to parse both files completely into later unused maps

If you haven't already done so, please make sure to sign the [CLA](https://github.com/jitsi/jitsi/blob/master/CONTRIBUTING.md#contributor-license-agreement).

networkinterfacecontroller:
Okay, makes sense to only parse for this particular property in ConfigurationActivator, I wasn't sure how though. I'm not very familiar with Java or with Jitsi's code (this was the first time I've been part of the Java ecosystem) and I pretty much relied on copying the way it was happening in JdbcConfig. Could you offer some advice or propose a different solution? I think it is good to have a way to activate prop config and turn off hsql, it makes it way easier to force provisioning with Active Directory (just replace a file on Login with the prov URI and prov mandatory set in default-overrides)

networkinterfacecontroller:
Seems like that was the intended function of usePropFileConfigService(), since per this codeblock and the windows patch notes for 2.10 adding ~/.usepropfileconfig would activate this feature in Linux.

ibauersachs:
Can you please explain your use case with Active Directory and why deploying the .msi with provisioning parameters doesn't work?
If we're talking about a Windows only situation, we could consider implementing a GPO based solution.

networkinterfacecontroller:
I hadn't thought of doing it that way, that would probably work for our use case. I still think it could be good to be able to use default-overrides, you can set options in case of provisioning failure, more options/flexibility in use.

ibauersachs:
If you want the hold thing reviewed, please open a separate pull request.

networkinterfacecontroller:
I did submit it as a separate pull request but for some reason it did two notifications, one as this pull and one as the new one. new one is #437

damencho:
This PR is not from a branch, but from the master and you pushed commits from both PRs in master. So the commit from the other PR polluted this PR ... that's why Ingo is commenting...

networkinterfacecontroller:
I did submit it as a separate pull request but for some reason it did two notifications, one as this pull and one as the new one. new one is #437

PR #436, opened 2018-01-11
fix jitsi could not transfer file with thumbnail

dmitrii-88:
The reason is that in the method **FileElement::parseIQ** where is the line <tt>thumbnail = new ThumbnailElement(parser.getText())</tt> where <tt>parser.getText()</tt> equals something like `<thumbnail xmlns="..." cid="..." mime-type="image/png" width="64" height="64">` (note closing '>' instead of '/>'). That's why exception occurs in the constructor of ThumbnailElement
    org.xml.sax.SAXParseException; lineNumber: 1; columnNumber: 160; XML document structures must start and end within the same entity.

### DIFF
--- a/src/net/java/sip/communicator/impl/protocol/jabber/extensions/thumbnail/FileElement.java
+++ b/src/net/java/sip/communicator/impl/protocol/jabber/extensions/thumbnail/FileElement.java
@@ -263,7 +263,7 @@ public class FileElement
                 }
                 else if (elementName.equals("thumbnail"))
                 {
-                    thumbnail = new ThumbnailElement(parser.getText());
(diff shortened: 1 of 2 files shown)


ibauersachs:
Thank you!
Before I can consider merging this, could you please make sure that you signed the [CLA](https://github.com/jitsi/jitsi/blob/master/CONTRIBUTING.md#contributor-license-agreement)?

dmitrii-88:
I sign CLA

ibauersachs:
Thanks! I've removed the link to the PDF to hide your personal information.

I'll target merging this once our OpenH264 stuff is through.

dmitrii-88:
Ok. Thanks.

PR #437, opened 2018-01-16
added support for playing long hold alerts

networkinterfacecontroller:
I added in a feature that checks for a property "net.java.sip.communicator.imp.protocol.sip.HOLD_ALERT_TIME". If this property is set, then Jitsi will play an alert noise when someone has been on hold for too long. This is really useful in a callcenter setting.
If the property is not set, the functionality is disabled.


networkinterfacecontroller:
Okay cool, this should be the only one open now and only includes changes related to the hold

networkinterfacecontroller:
This has a bug that will keep playing a sound after a call ends if it was transferred while someone was on hold. closing for now until it is patched

PR #449, opened 2018-02-17
Improve handling of OAuth2 code acquisition

### LINKED ISSUE #444: Google Contacts not working (anymore?)
A few days ago I installed Jitsi and I am having the same problems as already explained in this issue: #323 . However, when I read the articles it looks like it was solved and after that the discussion was about Fritz!Box. So I decided to create a new issue.

As with the old issue: With each letter that I enter there pops up a window and asks me for the auth code from google. I could successfully retrieve the authorization code from Google but it seems that the code is not being saved. This also happens if I run Jitsi as an Administrator.

This is my environment:
Jitsi 2.10.5550
Java 1.8.0_151
Windows 10 (1709 with all updates applied)

image

ibauersachs:
Closes #444


ibauersachs:
Wait for merge until OpenH.264 is merged.

PR #452, opened 2018-03-01
Adds addOctoRelays to ColibriBuilder.

bgrozev:


src/net/java/sip/communicator/impl/protocol/jabber/extensions/colibri/ColibriBuilder.java
```diff
@@ -1149,103 +1212,69 @@ public void setSimulcastMode(SimulcastMode simulcastMode)
...
+     * @param channel the channel to match.
      *
      * @return the channel from {@link #request} with the specified content
-     * name and ID.
+     * name and ID and run-time type matching this of {@code channel}.
      */
-    private ColibriConferenceIQ.ChannelCommon getRequestChannel(
-            String contentName, String channelId, boolean sctpConnection)
+    private <T extends ColibriConferenceIQ.ChannelCommon> T getRequestChannel(
+        String contentName, T channel)
```
bbaldino: it's a bummer we have to pass the whole thing in instead of just the id.  i was thinking we could pass the id and a .class and use reflection to achieve this (it would also remove the need for the if/else statement), but maybe that's going to far to try and avoid passing the whole object.

src/net/java/sip/communicator/impl/protocol/jabber/extensions/colibri/ColibriBuilder.java
```diff
@@ -782,14 +787,72 @@ public boolean addSourceInfo(
...
+        Objects.requireNonNull(localChannelsInfo, "localChannelsInfo");
+
+        if (conferenceState == null
+            || StringUtils.isNullOrEmpty(conferenceState.getID()))
+        {
+            // We are not initialized yet
+            return false;
+        }
+
+        assertRequestType(RequestType.CHANNEL_INFO_UPDATE);
```
bbaldino: i know this isn't new code, but i don't get what this is doing?  what is it asserting on/against?

src/net/java/sip/communicator/impl/protocol/jabber/extensions/colibri/ColibriBuilder.java
bgrozev: I mostly agree, but since it's a private method and we have the ChannelCommon instance everywhere where we need it, I thought that was clean enough

src/net/java/sip/communicator/impl/protocol/jabber/extensions/colibri/ColibriBuilder.java
bgrozev: There are two main types of requests: requests for new channels (ALLOCATE_REQUEST) and updating state for existing channels (CHANNEL_INFO_UPDATE). The API doesn't allow to mix the two, e.g. if we are building an allocate request, we don't know the IDs of the channels yet, so we can not use the addOctoRelays method (since it wouldn't know how to identify the channels to add relays to).

bbaldino: ok, i looked up the method and see it compares it against a member variable, and actually even re-assigns that member variable in `assertRequestType` 😮 .  not a big fan of the naming/semantics of that but i know it was already there so don't worry about it for this.

PR #462, opened 2018-04-05
Introduces openh264.

damencho:


resources/install/build.xml
```diff
@@ -2468,7 +2460,7 @@
                  value="$(PACKAGE_NAME)"/>
         <replace file="${debianize.dir}/rules"
                  token=" usr/lib/jni/"
-                 value=" usr/share/$(PACKAGE_NAME)/lib/native -Xslickless.jar -Xslick-runner.jar -X-slick.jar -X.svn"/>
+                 value=" usr/share/$(PACKAGE_NAME)/lib/native/"/>
```
ibauersachs: Why stop excluding the slick stuff?

resources/install/debian/rules.tmpl
```diff
@@ -31,6 +37,16 @@ ifeq ($(DEB_HOST_ARCH),amd64)
 else
 	dh_install -p$(PACKAGE_NAME)-jni lib/native/linux/*.so usr/lib/jni/
 endif
+	test -e $(USR_LIB_FOLDER)/libavformat-ffmpeg.so.56 && dh_link $(USR_LIB_FOLDER)/libavformat-ffmpeg.so.56 usr/lib/jni/libavformat.so || true
```
ibauersachs: I assume usr/lib/jni is local to Jitsi and not the global /usr/lib? Otherwise this might create a huge mess...

resources/install/debian/rules.tmpl
```diff
@@ -31,6 +37,16 @@ ifeq ($(DEB_HOST_ARCH),amd64)
...
+	test -e $(USR_LIB_FOLDER)/libavcodec-ffmpeg.so.56 && dh_link $(USR_LIB_FOLDER)/libavcodec-ffmpeg.so.56 usr/lib/jni/libavcodec.so || true
+	test -e $(USR_LIB_FOLDER)/libavfilter-ffmpeg.so.5 && dh_link $(USR_LIB_FOLDER)/libavfilter-ffmpeg.so.5 usr/lib/jni/libavfilter.so || true
+	test -e $(USR_LIB_FOLDER)/libavutil-ffmpeg.so.54 && dh_link $(USR_LIB_FOLDER)/libavutil-ffmpeg.so.54 usr/lib/jni/libavutil.so || true
+	test -e $(USR_LIB_FOLDER)/libswscale-ffmpeg.so.3 && dh_link $(USR_LIB_FOLDER)/libswscale-ffmpeg.so.3 usr/lib/jni/libswscale.so || true
+	test -e $(USR_LIB_FOLDER)/libavformat.so.57 && dh_link $(USR_LIB_FOLDER)/libavformat.so.57 usr/lib/jni/libavformat.so || true
```
ibauersachs: The non -ffmpeg links shouldn't be necessary since our stripped libjnffmpeg.so doesn't link to a specific version of ffmpeg and hence doesn't care about the binary version number.

resources/install/build.xml
damencho: Cause it is already there and this is duplication:
https://github.com/jitsi/jitsi/blob/master/resources/install/debian/rules.tmpl#L22

resources/install/debian/rules.tmpl
damencho: We are doing a replace for the actual path:
https://github.com/jitsi/jitsi/blob/07df10c510c1a23b30cf31c2d4f9db1a9cb23b65/resources/install/build.xml#L2461

This is because of the source package we used to build for debian official release. Maybe later we can get rid of all that mess, as nobody will be working on that and anyway this is not the right path for it.

resources/install/build.xml
ibauersachs: Not sure I get all this Debian crap 🤷. Hope we can build a proper package and get rid of it at some point.

resources/install/debian/rules.tmpl
damencho: Hum, but the binaries has this:
libavformat.so => /usr/lib/x86_64-linux-gnu/libavformat.so
And there is no such thing on the systems with libavformat57 installed?

resources/install/debian/rules.tmpl
ibauersachs: Ah, the template 🤦‍♂️

resources/install/debian/rules.tmpl
damencho:
```
# ls -l /usr/lib/x86_64-linux-gnu/ | grep format
lrwxrwxrwx  1 root root       24 Sep 26  2017 libavformat.so.57 -> libavformat.so.57.71.100
-rw-r--r--  1 root root  2387864 Sep 26  2017 libavformat.so.57.71.100
```
This is on Ubuntu 17.10, do you think this is ok and we do not need the symlinks?

damencho: Yep, I just tested that on Ubuntu 17.10 where libavformat57 is installed.
```
# ldd libjnffmpeg.so
	linux-vdso.so.1 =>  (0x00007ffc759a6000)
	libavformat.so => not found
```

But adding a symlink as in the rules file:
```
/tmp# LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/tmp ldd libjnffmpeg.so
./libjnffmpeg.so: libavformat.so: version `LIBAVFORMAT_FFMPEG_56' not found (required by ./libjnffmpeg.so)
	linux-vdso.so.1 =>  (0x00007ffc22d30000)
	libavformat.so (0x00007ff9ac536000)
```
It resolves it, but this 'not found' makes me think it will not work.

ibauersachs: The libjnffmpeg.so that I have still lying around works with ldd, your's doesn't. Both are x64 and readelf -d shows that they were compiled on a Ubuntu with the -ffmpeg versions. However, mine resolves libavformat.so, since I have a symlink /usr/share/x86_64-linux-gnu/libavformat.so -> /usr/share/x86_64-linux-gnu/libavformat-ffmpeg.so. However, this is from the [libavformat-dev package](https://packages.ubuntu.com/xenial/amd64/libavformat-dev/filelist).

Bummer. Seems there are two options:
- Use OpenH264 on Linux too
- Distribute two libjnffmpeg builds (one against -ffmpeg, one without the suffix, skip patchelf entirely)

damencho: I will try with having two binaries, and drop the patchelf thing.

damencho:
@ibauersachs I just force pushed the last to commits with latest changes in jitsi-lgpl-deps.

PR #468, opened 2018-04-18
Live stream url

bbaldino:
Adds support for passing the youtube broadcast id when starting a live stream so that it can be used to construct the url for viewers to see the stream.


bbaldino:
It may be tempting to make this field generic rather than YouTube specific, but given that we don't have other services planned yet (and therefore don't even know what their equivalent is to YouTube's broadcast ID is, if there is one at all), I think we should wait until we add another service before worrying about being able to re-use this field.

PR #477, opened 2018-05-14
Avoid NPE on packet logging when using xmpp bosh connection.

damencho:


ibauersachs:
How does this prevent an NPE?

damencho:
When connecting using bosh there is no socket, so we cannot obtain the address and we pass null as local and remote address and packet logger throws NPE.
There was a case handling that (new byte[4]), but not for all cases ...

src/net/java/sip/communicator/impl/protocol/jabber/debugger/SmackPacketDebugger.java
```diff
@@ -131,9 +125,9 @@ public void processStanza(Stanza packet)
 
                     packetLogging.logPacket(
                             PacketLoggingService.ProtocolName.JABBER,
-                            localAddress,
+                            localAddress != null ? localAddress : new byte[4],
```
ibauersachs: Since you now added the null check inside logPacket, these checks here are unnecessary.

damencho: Yes, I just force pushed.

PR #478, opened 2018-05-14
Pass username on xmpp conn.login.

damencho:
We need to pass username, not the bare jid when we call connection.login
.

### DIFF
--- a/src/net/java/sip/communicator/impl/protocol/jabber/LoginByPasswordStrategy.java
+++ b/src/net/java/sip/communicator/impl/protocol/jabber/LoginByPasswordStrategy.java
@@ -96,7 +96,8 @@ public class LoginByPasswordStrategy
     public boolean login(AbstractXMPPConnection connection, EntityFullJid jid)
         throws XMPPException, InterruptedException, IOException, SmackException
     {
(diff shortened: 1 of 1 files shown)


ibauersachs:
Are you sure about this? Isn't the domain part of the JID used to determine the realm on the server? I'd say that e.g. a GMail-Account wouldn't work anymore with this.

damencho:
I was testing jigasi smack-4-2 branch and it I was getting an error "Invalid username", then I turned on debug logging and saw an error in prosody that username contains invalid characters, then going to this method I saw that before smack upgrade it was getting a String param named username and the change from this PR fixed it connecting. 
https://github.com/jitsi/jitsi/commit/058d437f496701336fc9db722b8610b8eae124ec#diff-dc598bbb99b220ebffd71d91c77fde99L119
Jicofo works, cause it is setting AUTH_USERNAME or something like that. But this wasn't the case before for jigasi, but it was still connecting, so I thought we better fix that. I'm not sure about gmail, though... Haven't tested it...

ibauersachs:
String username didn't mean much in pre-Smack4. I'll see if I can test with a Google account tonight, unless you beat me to it.

PR #479, opened 2018-05-15
Fix usage of Certificate Verification Service

ibauersachs:


src/net/java/sip/communicator/impl/protocol/jabber/ProtocolProviderServiceJabberImpl.java
```diff
@@ -1195,6 +1181,16 @@ private ConnectState connectAndLogin(
                 }
 
                 confConn.setCustomSSLContext(sslContext);
+                confConn.setHostnameVerifier(new HostnameVerifier()
+                {
```
damencho: Why do we need this?

src/net/java/sip/communicator/impl/protocol/jabber/ProtocolProviderServiceJabberImpl.java
```diff
@@ -1219,6 +1230,7 @@ else if (loginStrategy.isTlsRequired())
             connection.addPacketInterceptor(debugger.outbound, null);
         }
 
+        connection.setReplyTimeout(30000);
```
damencho: Wasn't there already a timeout? Or it was in smack 3, and wasn't the timeout something like 5 or 10 seconds?

ibauersachs: Because Smack sets a default verifier which trows in case of a name mismatch:
https://github.com/igniterealtime/Smack/blob/master/smack-tcp/src/main/java/org/jivesoftware/smack/tcp/XMPPTCPConnection.java#L812

We don't want that, our verification service already validates the name (and allows to override a mismatch with the gui).

src/net/java/sip/communicator/impl/protocol/jabber/ProtocolProviderServiceJabberImpl.java
ibauersachs: Because Smack sets a default verifier which trows in case of a name mismatch:
https://github.com/igniterealtime/Smack/blob/master/smack-tcp/src/main/java/org/jivesoftware/smack/tcp/XMPPTCPConnection.java#L812

We don't want that, our verification service already validates the name (and allows to override a mismatch with the gui).

src/net/java/sip/communicator/impl/protocol/jabber/ProtocolProviderServiceJabberImpl.java
ibauersachs: And I even had timeouts without any validation failures. Granted, I was on a VPN, but both sides with >= 100 MBit/s

PR #481, opened 2018-05-16
Fix construction of (anonymous) BOSH connections and enable stats provider

damencho:
Some small changes to be able to run successfully jigasi.

### DIFF
--- a/src/net/java/sip/communicator/impl/protocol/jabber/ProtocolProviderServiceJabberImpl.java
+++ b/src/net/java/sip/communicator/impl/protocol/jabber/ProtocolProviderServiceJabberImpl.java
@@ -1114,8 +1114,40 @@ public class ProtocolProviderServiceJabberImpl
         confConn.setXmppDomain(serviceName);
         if (isBosh)
         {
-            confConn.setHost(boshURL);
-            // FIXME use setHostAddress, use proxy
+            BOSHConfiguration.Builder boshConfConnBuilder =
+                (BOSHConfiguration.Builder)confConn;
+
+            try
+            {
+                URI boshURI = new URI(boshURL);
+                boolean useHttps = boshURI.getScheme().equals("https");
+
+                int port = boshURI.getPort();
+                if (port == -1)
+                {
+                    port = useHttps ? 443 : 80;
+                }
+
+                String file = boshURI.getPath();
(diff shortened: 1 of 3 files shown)


src/net/java/sip/communicator/impl/protocol/jabber/ProtocolProviderServiceJabberImpl.java
```diff
@@ -1114,8 +1114,33 @@ private ConnectState connectAndLogin(
...
+
+                int port = boshURI.getPort();
+                if (port == -1)
+                {
+                    port = useHttps ? 443 : 80;
+                }
+
+                boshConfConnBuilder
+                    .setUseHttps(useHttps)
+                    .setFile(boshURI.getPath() + "?" + boshURI.getQuery())
```
ibauersachs: Is there always a query? I know it works with an empty string after the ?, but it's ugly.

damencho: I've updated and forced pushed the last commit. Does it look ok now @ibauersachs? Thanks.

src/net/java/sip/communicator/impl/protocol/jabber/ProtocolProviderServiceJabberImpl.java
```diff
@@ -1114,8 +1114,40 @@ private ConnectState connectAndLogin(
...
+
+                int port = boshURI.getPort();
+                if (port == -1)
+                {
+                    port = useHttps ? 443 : 80;
+                }
+
+                String file = boshURI.getPath();
+                String query = boshURI.getQuery();
+                if (!StringUtils.isNullOrEmpty())
```
ibauersachs: Missing argument? ;-)
I'm surprised that this even compiles.

damencho: Shit, sorry, hum...

damencho: Indeed it was compiling, because of isNotEmpty(CharSequence... css).
Fixed now.

damencho:
Thank you! :)

PR #483, opened 2018-05-19
 add ability to intercept Presence objects received by ChatRoomJabberImpl and add presence extensions

nikvaessen:


src/net/java/sip/communicator/impl/protocol/jabber/ChatRoomJabberImpl.java
```diff
@@ -3187,6 +3194,72 @@ private void processOtherPresence(Presence presence)
...
+
+    /**
+     * Enables adding extensions to the presence of the {@link ChatRoom} without
+     * explicitly adding getters and setters for the values stored in these
+     * extensions but instead giving a {@link ChatRoomPresenceListenerExtender}
+     * which handles the logic of reading the values and storing them in the
+     * {@link ChatRoomMemberJabberImpl} by using for example
+     * {@link ChatRoomMemberJabberImpl#setValue(String, Object)}
+     */
+    public interface ChatRoomPresenceListenerExtender
```
ibauersachs: Please put this into a separate file, i.e. make it a top-level class (interface).

src/net/java/sip/communicator/impl/protocol/jabber/ChatRoomJabberImpl.java
```diff
@@ -3187,6 +3194,72 @@ private void processOtherPresence(Presence presence)
...
+    {
+
+        /**
+         * Read a {@link Presence} and retrieve the associated
+         * {@link ChatRoomMemberJabberImpl} to potentially modify
+         *
+         * @param presence the presence
+         * @param member the member associated with the presence
+         */
+        void readPresenceAndModifyMember(Presence presence,
```
ibauersachs: The `read` prefix is awkward. Maybe just `presenceUpdated` or `memberPresenceUpdated`?

src/net/java/sip/communicator/impl/protocol/jabber/ChatRoomMemberJabberImpl.java
```diff
@@ -87,6 +89,14 @@
      */
     private String displayName;
 
+    /**
+     * A map to store {@link Object}s. Along with the provided methods
+     * {@link this#setValue(String, Object)} and {@link this#getValue(String)},
+     * this map prevents adding a getter/setter for each value which we want to
+     * associate with this {@link ChatRoomMember}.
+     */
+    private final Map<String, Object> valueMap = new HashMap<>();
```
ibauersachs: What is supposed to be stored here?

nikvaessen: The avatar-url, avatar-id or gravatar hash based on whatever is present, as well as group id and user-id from Stride organisation when deployed in that environment. Because there are so many different extensions used, it seemed better to add something like this instead of using a setter/getter for everything

bgrozev:
I don't like adding additional interfaces and listeners just for this. We already have ChatRoomMemberPresenceListener and ChatRoomMemberPresenceChangeEvent. Perhaps we can save and expose the last received Presence in ChatRoomMemberJabberImpl and use the existing listeners and interfaces?

src/net/java/sip/communicator/impl/protocol/jabber/ChatRoomMemberJabberImpl.java
ibauersachs: I'm not a fan of such maps but I can see the reasoning behind. @damencho ?

damencho: I also see the reason behind that map. But we already have such mechanism for calls, @nikvaessen you can follow the same pattern. Just make ChatRoomMemberJabberImpl extend DataObject, where you will have getData(Object key) and setData(Object key, Object value). 

The other thing here is the newly introduced interface which exposes smack object in its interface, something we were trying to avoid. There is already one such precedent in OperationSetJitsiMeetTools... but, if we can avoid adding more ...

So you can add ChatRoomMemberPresenceChangeEvent.MEMBER_UPDATED and reuse ChatRoomMemberPresenceListener for your needs (as Boris proposed). And you can store lastReceivedPresence and add a method getLastPresence for ChatRoomMemberJabberImpl and handle storing of data that way, avoiding exposing smack API in the jitsi protocol providers API. 
Does this sound ok?

src/net/java/sip/communicator/impl/protocol/jabber/ChatRoomMemberJabberImpl.java
```diff
@@ -32,6 +35,7 @@
  * @author Emil Ivov
  */
 public class ChatRoomMemberJabberImpl
+    extends DataObject
```
bgrozev: Do we still need that? I don't see any of the `DataObject` methods being used.

damencho:
I will need this for my changes in jigasi :)

src/net/java/sip/communicator/impl/protocol/jabber/ChatRoomMemberJabberImpl.java
damencho: Yep, it is needed (we synced up offline with Nik and Boris)

PR #487, opened 2018-05-25
Workaround for HiDPI support on Linux.

gpolitis:
It looks like the GtkLookAndFeel won't play nice with HiDPI graphics:

https://bugs.openjdk.java.net/browse/JDK-8055212
https://bugs.openjdk.java.net/browse/JDK-8058742

I don't know what the best approach to permanently fix this, but a
workaround is to use Java >= 9 that is HiDPI aware and the Motif look
and feel.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -872,6 +872,7 @@
             <sysproperty key="user.language" value="${user.language}" />
             <sysproperty key="user.country" value="${user.country}" />
             <sysproperty key="user.variant" value="${user.variant}" />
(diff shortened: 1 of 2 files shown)


ibauersachs:
Java 9 is already EOL, Apache Felix doesn't run on Java 10/11 yet, and statically setting the scale to 2 seems wrong for non-HiDPI screens.

gpolitis:
I'm sorry, I should've explicitly said so, but please don't merge this, it just provides a workaround and it can serve as a basis for discussion!

PR #491, opened 2018-06-12
Additions to ColibriStatsExtension

bgrozev:


src/net/java/sip/communicator/impl/protocol/jabber/extensions/colibri/ColibriStatsExtension.java
```diff
@@ -45,6 +46,25 @@
...
+     * @param source the {@link ColibriStatsExtension} to copy.
+     * @return the copy.
+     */
+    public static ColibriStatsExtension clone(
+        ColibriStatsExtension source)
+    {
+        ColibriStatsExtension destination
+            = AbstractPacketExtension.clone(source);
+
+        for (Stat stat : source.getChildExtensionsOfType(Stat.class))
```
bbaldino: it looks like child extensions are known to `AbstractPacketExtension`, so couldn't it handle cloning all of them in its clone method?

bgrozev: The child extensions are of type `? extends ExtensionElement`, so they can not always be cloned with `AbstractPacketExtension.clone()`.  I guess this could be worked around somehow, but it would require more work.

bbaldino: oh, i assumed the extension elements had a clone method defined...that's unfortunate, this means we'd have to remember to implement clone in all the extensions which may have child elements when it could've been done in one place.  too bad.  ok well don't need to take care of it here then.

PR #498, opened 2018-06-18
Remove namespace from ContentPacketExtension

jahndis:
This commit sets the namespace of the ContentPacketExtension to null. This was causing issues when parsing the jingle session-initiate message with stanza.io. The namespace was being set on content elements inside a group which stanza.io would fail to parse.

According the group XEP there usually isn't a namespace on the content elements (https://xmpp.org/extensions/xep-0338.html), and the namespace currently doesn't show up on the content elements inside the jingle element in the session-initiate message. (smack 4.2 might be doing something smart to remove a child element's namespace if it is the same as the parent).

More details in this discussion: https://community.jitsi.org/t/jitsi-dev-custom-client-fails-to-parse-jingle-session-initiate-with-new-jicofo/13992

Ran tests in jitsi-videobridge, jicofo, jicoco, and jigasi and all passed.

### DIFF
--- a/src/net/java/sip/communicator/impl/protocol/jabber/extensions/jingle/ContentPacketExtension.java
+++ b/src/net/java/sip/communicator/impl/protocol/jabber/extensions/jingle/ContentPacketExtension.java
@@ -37,7 +37,7 @@ public class ContentPacketExtension extends AbstractPacketExtension
     /**
      * The namespace of the "content" element
      */
-    public static final String NAMESPACE = "urn:xmpp:jingle:1";
+    public static final String NAMESPACE = null;
 
     /**
      * The name of the "creator" argument.


damencho:
@bgrozev @bbaldino WDYT? 
I think this is safe and this was the way it used to be prior smack4 update.

bbaldino:
so do we not need to set the namespace based on its usage? (i.e. ingo's comment [here](https://community.jitsi.org/t/jitsi-dev-custom-client-fails-to-parse-jingle-session-initiate-with-new-jicofo/13992/4))

damencho:
Yep, I'm not sure about that. 
But if smack is removing namespace when they match the parent as @jahndis reported,  
then defining two namespaces in ContentPacketExtension and make namespace as a required constructor parameter and passing the correct one depending on the usage will be a better approach. WDYT? 
@jahndis can you try that?

bgrozev:
How does this affect parsing?

jahndis:
Sorry for the delay in response. I was sick for a couple days and then on vacation through the rest of the week.

I made the changes you recommended, and reran the tests (all passed). Let me know if there's anything else needed.

PR #505, opened 2018-07-13
doc: Updates the debian copyright file, remove mention of files we ar…

bgrozev:
…e no longer using or distributing.


ibauersachs:
Ahm. What's going on here? When was the Thunderbird integration removed (jmork)? Spelling in the chat window? IRC?
If this is something for Meet, please add a different copyright file in another directory.

Issue #135, opened 2015-07-05
OSX Upgrade allowed even when not


Aviator45003:
When on OSX, Jitsi allows an upgrade to a version which then can not run on that version of OSX. Simple version check would be great.

ibauersachs:
What versions of OSX are we talking about?

Aviator45003:
10.8.2.

ibauersachs:
We do support 10.8, however you should upgrade to at least 10.8.5 for security reasons. Your short messages aren't really helpful in finding the problem. We'd need a detailed error description, the versions of Jitsi you used, the error message(s) and your logs.

Issue #245, opened 2016-03-29
Desktop Sharing Linux One Way Only


BoBeR182:
I am running Arch linux and using the AUR for jitsi [0]

I can receive a shared desktop from a Windows user, and if they allow remote control I can control it.

If I share my desktop, fixed region or full-screen, they do not see anything or are able to control anything.

What would help debug this?

[0] https://aur.archlinux.org/packages/jitsi/

ibauersachs:
Reading the bug reporting guidelines would a start. They tell you to write to the mailing list and send the [logs](https://jitsi.org/logs).

BoBeR182:
Thank you for the direction, I am slightly confused.
It lists to send my bug to the dev list here [0], and to send my bug to the users list here [1].

[0] https://jitsi.org/Documentation/ReportingBugs
[1] https://jitsi.org/Development/BugsAndIssues

GNUDimarik:
Did anyone try to reproduce it under Ubuntu?

damencho:
@GNUDimarik libjnffmpeg is from here: https://github.com/jitsi/jitsi-lgpl-dependencies
The rest are from here: https://github.com/jitsi/libjitsi

GNUDimarik:
@damencho thanks!

GNUDimarik:
@damencho please let me know:
libjnffmpeg.so
libjnffmpeg-ffmpeg.so
What's difference?

When I built native libraries it built 2 libffmpeg.so with and without h264 support.
When I built new ones with statically built ffmpeg with h264 support it started to work. The both: camera and screen sharing. I think it was because I had irrelevant libav* libraries and static likning should help us to avoid it.
Is that right way? If so
will do the same for linux 32 bit and make pull request to https://github.com/jitsi/jitsi-lgpl-dependencies

GNUDimarik:
Here is dependencies list:
ldd jnffmpeg_oh264/libjnffmpeg_oh264.so 
	linux-vdso.so.1 =>  (0x00007fffa71a2000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f9d44e52000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f9d44a89000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f9d4553a000)
I think it's good enough for making it more portability. I'm sure this library should work on most linux distors but of course I tested it locally only.
[...]

PR #519, opened 2018-08-10
Fixes #104. Now it's possible to type on remote machine and works sim…

### LINKED ISSUE #104: (Keymap?) issue in Jitsi remote screen
I discussed this on the Jitsi mailing lists and the link to that discussion is: http://lists.jitsi.org/pipermail/users/2014-February/006394.html

I am seeing an issue with typing in a remote screen session over Jitsi. While I can see the screen of the other party, move the mouse and make simple inputs, special keys do not work properly. In particular, pressing backspace or enter in a terminal window or using the backspace key is broken.
- Backspace triggers H (Ctrl-H),
- Enter triggers J (Ctrl-J)

and I see both keystrokes as h and j in the terminal, telling me the Ctrl switch was not transferred.

On a side note, pressing Enter (triggering j in the remote terminal) switches to the previous tab in the local Jitsi ;) ...

(Issue migrated from https://trac.jitsi.org/ticket/1277)

GNUDimarik:
Now it's possible to type on remote machine and works simple navigation such as backspace, enter and arrow keys.
Please review and let me know your opinions.
Sorry I'm tired after working day and before I made pull request in wrong issue (#135).
This one has right target

### DIFF
--- a/src/net/java/sip/communicator/impl/gui/main/call/DTMFHandler.java
+++ b/src/net/java/sip/communicator/impl/gui/main/call/DTMFHandler.java
@@ -570,6 +570,7 @@ public class DTMFHandler
     private void startSendingDtmfTone(Call call, DTMFToneInfo info)
     {
         Iterator<? extends CallPeer> callPeers = call.getCallPeers();
+        boolean notificationAdded = false;
 
         try
         {
@@ -589,8 +590,24 @@ public class DTMFHandler
                             .getCurrentCallRenderer()
                                 .getCallPeerRenderer(peer);
 
-                    if (peerRenderer != null)
+                    if (peerRenderer != null && peerRenderer.isDtmfToneEnabled())
+                    {
+                        if (!notificationAdded)
+                        {
+                            if(info.sound != null)
+                            {
+                                synchronized(dtmfToneNotifications)
+                                {
+                                    dtmfToneNotifications.add(info);
+                                    startDTMFToneNotificationThreadIfNecessary();
+                                }
+                            }
+
+                            notificationAdded = true;
+                        }
+
                         peerRenderer.printDTMFTone(info.keyChar);
+                    }
                 }
             }
         }
@@ -612,15 +629,6 @@ public class DTMFHandler
      */
     private synchronized void startSendingDtmfTone(DTMFToneInfo info)
     {
-        if(info.sound != null)
-        {
-            synchronized(dtmfToneNotifications)
-            {
-                dtmfToneNotifications.add(info);
-                startDTMFToneNotificationThreadIfNecessary();
-            }
-        }
-
         Collection<Call> calls
             = (callContainer == null)
                 ? CallManager.getInProgressCalls()
--- a/src/net/java/sip/communicator/impl/gui/main/call/DesktopSharingMouseAndKeyboardListener.java
+++ b/src/net/java/sip/communicator/impl/gui/main/call/DesktopSharingMouseAndKeyboardListener.java
@@ -17,6 +17,7 @@
  */
 package net.java.sip.communicator.impl.gui.main.call;
 
+import net.java.sip.communicator.service.gui.call.CallPeerRenderer;
 import net.java.sip.communicator.service.protocol.*;
 import net.java.sip.communicator.service.protocol.event.*;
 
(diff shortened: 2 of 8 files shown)


GNUDimarik:
Will update the pull request soon since I have working better fix.
But I have also one question: keyboard events was grabbed because DTMFHandler tries to send us DTMF code and now I ignore this in renderer which can know when remote control on other side is enabled. But sound is still plays. From other side this is right place because we can have someone else on the call and that peer should handle DTMF since he doesn't enable remote control (as example) on his side.
So what do you think current solution is acceptable at least as start point?
Will provide the code a little later when I'll have feedback from QA.
Just I think deep testing may show us additional issues ...

src/net/java/sip/communicator/impl/gui/main/call/DesktopSharingMouseAndKeyboardListener.java
```diff
@@ -22,19 +22,22 @@
...
  */
 public class DesktopSharingMouseAndKeyboardListener
-    implements RemoteControlListener,
-               KeyListener,
-               MouseListener,
-               MouseMotionListener
+        implements RemoteControlListener,
+        KeyListener,
+        MouseListener,
+        MouseMotionListener
```
ibauersachs: Please avoid unnecessary changes

GNUDimarik: Ok. Thanks. WIll do

GNUDimarik:
Please don't merge this until I didn't update the commit

src/net/java/sip/communicator/impl/hid/HIDServiceImpl.java
```diff
@@ -77,19 +77,19 @@ protected HIDServiceImpl()
      */
     public void keyPress(int keycode)
     {
-        if(OSUtils.IS_WINDOWS || OSUtils.IS_MAC)
-        {
+        //if(OSUtils.IS_WINDOWS || OSUtils.IS_MAC)
```
GNUDimarik: I decided to comment this code because if we work with Ctrl, Shift and Alt like with modifiers only we can't handle hotkeys like Ctrl+Shif+Left(Right) etc

src/net/java/sip/communicator/impl/gui/main/call/DTMFHandler.java
```diff
@@ -227,6 +227,11 @@
      */
     public static final String DTMF_TONE_PREFIX = "DTMFTone.";
 
+    /**
+     * True if sending DTMF tone is enabled
+     */
+    private static boolean sendDTMFTone = true;
```
ibauersachs: This must be specific to a call and thus cannot be static.

src/net/java/sip/communicator/impl/gui/main/call/DesktopSharingMouseAndKeyboardListener.java
```diff
@@ -244,6 +250,7 @@ public void remoteControlGranted(RemoteControlGrantedEvent event)
         {
             if(videoComponent != null)
             {
+                DTMFHandler.enableSendingDTMFTone(false);
```
ibauersachs: You can enable/disable the DTMFHandler by adding methods to CallPeerRenderer and get the associated renderer via
`CallManager.getActiveCallContainer(getCallPeer().getCall()).getCurrentCallRenderer().getCallPeerRenderer(getCallPeer())` or you can pass it down into the constructor of `DesktopSharingMouseAndKeyboardListener`.

src/net/java/sip/communicator/impl/hid/HIDServiceImpl.java
ibauersachs: Did you test this on all platforms, i.e. Linux, Mac and Windows?
If the code isn't needed anymore, remove it instead of leaving it commente.d

GNUDimarik:
@ibauersachs 
Tested under Mac and it's not bad but I think cooperation between mac and other systems requires some additional work. Shift works fine and typing too. But when I got remote control on Mac from windows I expect Ctrl+C will copy to clipboard but now it doesn't work because Mac has another keys configuration. I think we need emulate this behaviour. And the same vise wersa. Mac user expects something from his kbd when he controlls windows machine. So I think we should finish this and next back to this pull request.

GNUDimarik:
Testing under linux will possible after fixing this issue: https://github.com/jitsi/jitsi/issues/245

GNUDimarik:
I got it. But anyway there should be way for making possible to conveniently work with remote control from windows to mac and wise versa. The both windows and mac are most popular platrorms and I think this case is real.
let's add a separate issue for that. What do you think?
By this issue: after testing under linux I'll fix the comments.

GNUDimarik:
Under linux the same behaviour like under windows. Tested between Ubuntu 16 and Windows 10

GNUDimarik:
@ibauersachs please check ammended commit

PR #523, opened 2018-08-23
Fixes #521 by adding sharing region before staring remote control and destroys origin when sharing is disabled

### LINKED ISSUE #521: Wrong mouse coordinates when remote control enabled during part screen sharing session
When you trying to share part of your screen you should enable remote control. After please try to mouse move and you will be able to notice cursor position is wrong.
Will make pull request with fix for this soon

GNUDimarik:
Here is the build for testing: https://www.dropbox.com/s/h3pj3qdcqno9u3v/jitsi-2.11.0-x86.exe?dl=0

### DIFF
--- a/src/net/java/sip/communicator/impl/gui/main/call/DesktopSharingFrame.java
+++ b/src/net/java/sip/communicator/impl/gui/main/call/DesktopSharingFrame.java
@@ -529,12 +529,16 @@ public class DesktopSharingFrame
                     frame.dispose();
 
                     if (call != null)
+                    {
                         CallManager.enableRegionDesktopSharing(
                                 call,
                                 location.x,
                                 location.y,
                                 sharingRegionWidth,
                                 sharingRegionHeight);
+
+                        setStreamingLocation(call, frame, location);
+                    }
                     else
                         CallManager.createRegionDesktopSharing(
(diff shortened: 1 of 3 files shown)


src/net/java/sip/communicator/impl/gui/main/call/DesktopSharingFrame.java
```diff
@@ -528,13 +528,16 @@ public void actionPerformed(ActionEvent e)
 
                     frame.dispose();
 
-                    if (call != null)
+                    if (call != null) {
```
ibauersachs: Formatting

src/net/java/sip/communicator/impl/gui/main/call/DesktopSharingFrame.java
```diff
@@ -249,7 +249,7 @@ private static void addDesktopSharingListener(final Call call, JFrame frame)
     {
         OperationSetVideoTelephony videoTelephony
             = call.getProtocolProvider().getOperationSet(
-                    OperationSetVideoTelephony.class);
+                OperationSetDesktopSharingServer.class);
```
ibauersachs: This change seems unneccessary.

Issue #530, opened 2018-09-14
jitsi not working with java 10


callegar:
Verified on ubuntu bionic.

## Description

when the java runtime environment is openjdk 10 (which is the default in bionic, even though bionic is expected to transition to openjdk 11 soon), jitsi fails.

---

## Current behavior

launching jitsi gets you the following error message:

```
Exception in thread "main" java.lang.NoSuchFieldError: preferIPv6Address
        at java.base/java.net.InetAddress.init(Native Method)
        at java.base/java.net.InetAddress.<clinit>(InetAddress.java:337)
        at java.base/java.net.NetworkInterface.init(Native Method)
        at java.base/java.net.NetworkInterface.<clinit>(NetworkInterface.java:68)
        at net.java.sip.communicator.util.launchutils.SipCommunicatorLock.findLocalAddress(SipCommunicatorLock.java:513)
        at net.java.sip.communicator.util.launchutils.SipCommunicatorLock.readLockFile(SipCommunicatorLock.java:413)
        at net.java.sip.communicator.util.launchutils.SipCommunicatorLock.readLockFileRetrying(SipCommunicatorLock.java:358)
        at net.java.sip.communicator.util.launchutils.SipCommunicatorLock.tryLock(SipCommunicatorLock.java:154)
        at net.java.sip.communicator.launcher.SIPCommunicator.main(SIPCommunicator.java:139)
```

---

## Expected Behavior

jitsi should either work fine on openjdk 10 or avoid to pick openjdk 10 as an acceptable jre.

---

## Possible Solution

I believe there are two possible solutions here.

1. This is the long term solution. Make jitsi work correctly with openjdk 10 and 11.
2. Interim solution (that should anyway be adopted IMHO). Modify the jitsi startup script.   Currently the shell script used to start jitsi includes the following item:
```
javabin=`which java`
```
This should be modified to let the jre be chosen via a shell variable (e.g. JITSI_JRE), defaulting to the one found by `which java` only if the shell variable is not configured. The rationale is the following. In the current situtation, to run jitsi, you need either to change the default jre for the whole of your system (via the ubuntu/debian alternatives mechanism) or manually edit the jitsi startup script. Either thing is inconvenient. The second one is inconvenient because whenever jitsi is updated, you need to re-edit the shell script. By having a shell variable, you could configure jitsi to use a suitable jre without touching the rest of your system.

---

## Steps to reproduce

[...]

PR #531, opened 2018-09-16
Let JRE be selected via shell variable on debian

callegar:
An issue with openjdk 10 prevents jitsi from running on debian derivatives (including ubuntu bionic) where the system is configured
to run the java binary from openjdk 10 (see bug #530).  Unfortunately, this is the default for ubuntu bionic.

Let the user decide which jre to use with jitsi via shell variable JITSI_JRE_HOME, so that it is possible to install Java 8 on ubuntu and use it with jitsi, without having to make the whole system default to Java 8 or without forcing the user to edit the jitsi script.

For instance, with this patch, one can say:

JITSI_JRE_HOME=/usr/lib/jvm/java-1.8.0-openjdk-amd64/jre jitsi

### DIFF
--- a/resources/install/debian/jitsi.sh.tmpl
+++ b/resources/install/debian/jitsi.sh.tmpl
@@ -25,7 +25,11 @@ if $show_splash ; then
     SPLASH_ARG="-splash:splash.gif"
 fi
 
-javabin=`which java`
+if [ -z "$JITSI_JRE_HOME" ] ; then
+    javabin=`which java`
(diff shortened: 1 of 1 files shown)


ibauersachs:
If you haven't already done so, could you please make sure you sign our CLA (https://jitsi.org/icla for individuals and https://jitsi.org/ccla for corporations)? We would unfortunately be unable to merge your patch unless we have that piece :(

PR #533, opened 2018-09-20
Fixes #75 by returning contact addres instetad of default toString() …

### LINKED ISSUE #75: Contact list filter does not correctly drop chat room member
Steps to reproduce:
- Join a Jabber chat room with members in it.
- Drag one of the members (I don't have her in my contact list) from the chat room chat dialog onto the contact list filter in the main window.

What happens:
The filter displays the text net.java.sip.communicator.impl.gui.main.chat.conference.ConferenceChatContact@c0dd841 instead of the id of the dropped chat room member

What is expected to happen:
The contact list filter in the main window should be filled with the Jabber ID of the dropped chat room member.

(Issue migrated from https://trac.jitsi.org/ticket/803)

GNUDimarik:
…implementation

### DIFF
--- a/src/net/java/sip/communicator/impl/gui/main/chat/conference/ConferenceChatContact.java
+++ b/src/net/java/sip/communicator/impl/gui/main/chat/conference/ConferenceChatContact.java
@@ -84,4 +84,15 @@ public class ConferenceChatContact
     {
         return descriptor.getContactAddress();
     }
+
+    @Override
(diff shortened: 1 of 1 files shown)


GNUDimarik:
@ibauersachs please review when you can

ibauersachs:
While looking around where the bug came from, I've implemented this differently in #534. Please have a look if this works for you.

PR #535, opened 2018-09-28
Irc ignore

Dunkhan:
This change gives the user an optional field in the advanced settings called "Irc ignore list".
If any nicks are written in here all incoming messages from these will be blocked.

It should be noted that the goal here was not to completely block these nicks but merely to prevent jitsi from conflating chanserv spam with real new messages from actual people. The solution implemented here is overzealous. A perfect solution would still record logs of the messages and also allow the user to initiate pm sessions with these nicks.

Implementing the ideal solution however requires changing far more files, many of which are outside the scope of the irc plugin. As a new committer and an inexperienced programmer I do not feel that it is a good idea for me to implement this as it is a restructuring of the whole messaging workflow.

As it stands this change saves a lot of annoyance for irc users, and is optional and easy to disable for situations where its effects are undesired.


ibauersachs:
If you haven't already done so, could you please make sure you sign our CLA (https://jitsi.org/icla for individuals and https://jitsi.org/ccla for corporations)? We would unfortunately be unable to merge your patch unless we have that piece :(

src/net/java/sip/communicator/impl/protocol/irc/irc.provider.manifest.mf
```diff
@@ -10,11 +10,13 @@ Import-Package: org.osgi.framework,
  org.jitsi.service.resources,
  net.java.sip.communicator.service.resources,
  net.java.sip.communicator.util,
+ net.java.sip.communicator.plugin.desktoputil,
```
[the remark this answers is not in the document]
Dunkhan: Do you mean I just have to move the UI stuff to another folder, or are more complex changes needed? Also is the general chat UI or the registration wizard preferred?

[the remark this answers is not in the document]
Dunkhan: I just noticed that question is answered in another comment below
I considered putting it there, and also in the general UI stuff but I was unsure.

ibauersachs:
The feature itself seems useful, I'll be one of the first users to block Freenode's annoying messages when returning from sleep. An independent block user/chatroom/bot feature would definitely be better, but also more complicated to implement. Up to you to give it a go.

As for your other question for the UI removal from the protocol: the ircaccregwiz package is a better option for the config panel. The protocol package should remain headless (i.e. able to run on servers).

src/net/java/sip/communicator/impl/protocol/irc/properties/IrcProperties.java
```diff
@@ -0,0 +1,15 @@
+package net.java.sip.communicator.impl.protocol.irc.properties;
+
+/**
+ * Properties for the IRC plugin
+ *
+ * @author Duncan Robertson
+ */
+public class IrcProperties
+{
+    public static final String PROP_IRC_IGNORE = "net.java.sip.communicator.impl.protocol.irc.properties.IGNORE_STRING";
```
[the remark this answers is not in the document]
Dunkhan: I am still unclear on this. I am not particularly attached to having a properties class. There is only one thing in it. I am not sure where you want me to put it though. What is meant by the protocol and the account assistant?

Dunkhan:
what do I do with signed CLA once I have it?

resources/languages/resources.properties
```diff
@@ -1031,6 +1031,9 @@ plugin.ircaccregwizz.SASL_USERNAME=Username
 plugin.ircaccregwizz.SASL_AUTHZ_ROLE=Role
 plugin.ircaccregwizz.SASL_IRC_PASSWORD_USED=IRC password above is used for SASL authentication
 plugin.ircaccregwizz.RESOLVE_DNS_THROUGH_PROXY=Always resolve DNS names through proxy
+plugin.irc.IRC_IGNORE_DESCRIPTION=<html> Add the names of contacts whose messages should not open a \
+new window or trigger notifications. e.g. chanserv</html>
+plugin.irc.IRC_IGNORE_CONFIG=Irc ignore list
```
ibauersachs: IRC is written uppercase erverywhere else

resources/languages/resources.properties
```diff
@@ -1766,4 +1769,4 @@ plugin.propertieseditor.NEED_RESTART=Note: some properties might need a restart
 #Thunderbird address book plugin
 plugin.thunderbird.CONFIG_FORM_TITLE=Thunderbird
 plugin.thunderbird.ENABLED=Enabled
-plugin.thunderbird.PHONE_PREFIX=Phone prefix:
+plugin.thunderbird.PHONE_PREFIX=Phone prefix:
```
ibauersachs: Files should end with a newline.

src/net/java/sip/communicator/impl/protocol/irc/irc.provider.manifest.mf
```diff
@@ -10,11 +10,11 @@ Import-Package: org.osgi.framework,
...
```
ibauersachs: The import shouldn't be here anymore.

src/net/java/sip/communicator/impl/protocol/irc/irc.provider.manifest.mf
```diff
@@ -10,11 +10,11 @@ Import-Package: org.osgi.framework,
...
  net.java.sip.communicator.service.certificate,
  net.java.sip.communicator.service.protocol,
  net.java.sip.communicator.service.protocol.event,
  org.slf4j,
- javax.net.ssl,
```
ibauersachs: Are you sure this isn't needed?

src/net/java/sip/communicator/impl/protocol/irc/properties/IrcProperties.java
```diff
@@ -0,0 +1,15 @@
...
+
+    public IrcProperties()
```
ibauersachs: This constructor is redundant. If at all, make it private to mark IrcProperties as a utility class.

src/net/java/sip/communicator/plugin/ircaccregwizz/IrcIgnoreConfigForm.java
```diff
@@ -0,0 +1,215 @@
+package net.java.sip.communicator.plugin.ircaccregwizz;
+
+import net.java.sip.communicator.impl.protocol.irc.properties.*;
```
ibauersachs: You'll need to copy the property. The ircaccregwiz is a different bundle and cannot use anthing from the protocol impl.

ibauersachs:
Damencho will check that they've received the CLA. You don't have to do anything else.

damencho:
I don't see the CLA in the list of signed ones, the last update was on Oct 09, 2018.
@Dunkhan have you submitted it to DocuSign?

PR #539, opened 2018-10-19
overhaul the jibri status presence extension

bbaldino:
jibri's status is now composed of 2 'sub statuses': its busy status
(whether or not this jibri is currently handling a request) and
its health status (whether or not this jibri has encountered and error
and needs intervention before it can handles requests again).

### DIFF
--- a/src/net/java/sip/communicator/impl/protocol/jabber/extensions/jibri/JibriStatusPacketExt.java
+++ b/src/net/java/sip/communicator/impl/protocol/jabber/extensions/jibri/JibriStatusPacketExt.java
@@ -63,22 +63,10 @@ public class JibriStatusPacketExt
         );
     }
 
-    private <T extends ExtensionElement> T getFirstChildExtensionOrNull(Class<T> type)
-    {
-        List<T> childExts = getChildExtensionsOfType(type);
-        if (!childExts.isEmpty())
-        {
-            return childExts.get(0);
-        }
-        else
-        {
-            return null;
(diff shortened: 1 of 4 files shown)


src/net/java/sip/communicator/impl/protocol/jabber/extensions/jibri/JibriStatusPacketExt.java
```diff
@@ -64,57 +63,48 @@ static public void registerExtensionProvider()
         );
     }
 
-    public Status getStatus()
+    private <T extends ExtensionElement> T getFirstChildExtensionOrNull(Class<T> type)
```
bgrozev: Should this go to AbstractPacketExtension? Seems generic and potentially useful

bbaldino: sure, maybe i should just call it `getChildExtension` (and keep this same behavior) to match the `setChildExtension` i added?

ibauersachs: Since this is Java 8+, instead of getFirstChildExtensionOrNull name this `Optional<T> getFirstChildExtension`. This is the _perfect_ use case for `Optional`s.

ibauersachs:
Pull requests are pointless if there isn't even time to review them...

Issue #562, opened 2019-01-18
Bouncy Castle 1.70 (2021-11-29) instead of 1.54 (CVEs)


Neustradamus:
@emcho, @damencho, @ibauersachs, @saghul: Urgently, can you update Bouncy Castle 1.54 to 1.70 (2021-11-29)?

PR for Jitsi: https://github.com/jitsi/jitsi/pull/564

- https://github.com/jitsi/jitsi-universe-public/blob/master/pom.xml#L37 (archived)

- https://github.com/jitsi/jitsi/blob/master/build.xml#L989
- https://github.com/jitsi/jitsi/blob/master/m2/pom.xml#L13

- https://github.com/jitsi/jitsi/blob/master/m2/jitsi-certificate/pom.xml#L21

- https://github.com/jitsi/webrtc/blob/master/DEPS#L230 (linked to Chromium)

- http://www.bouncycastle.org/latest_releases.html
- http://www.bouncycastle.org/releasenotes.html

CVEs:
- https://cve.mitre.org/cgi-bin/cvekey.cgi?keyword=bouncy%20castle
- https://cve.mitre.org/cgi-bin/cvekey.cgi?keyword=bouncycastle
- https://www.cvedetails.com/vulnerability-list/vendor_id-7637/Bouncycastle.html

Linked to: https://github.com/jitsi/jitsi-universe-public/issues/1

saghul:
Aare we affected by the issues described in the CVEs? At least one of them mentions that it only affects a certain low-level API usage.

Just trying to gauge the urgency here.

Neustradamus:
@saghul: There are a lot because there are very old Bouncy Castle versions in code...
- https://www.cvedetails.com/vulnerability-list/vendor_id-7637/Bouncycastle.html

And you can look on Jitsi community, problems and solutions:
- https://community.jitsi.org/search?q=bouncycastle%20order%3Alatest

After the little version updates in files, maybe good to see all other tickets about XMPP ^^

Thanks in advance!

Neustradamus:
Linked to https://github.com/jitsi/libjitsi/pull/461.

Neustradamus:
Updated to 1.61.

Neustradamus:
Updated to 1.63.
cc: @GNUDimarik.

Neustradamus:
@emcho, @damencho, @ibauersachs, @saghul: Last version is 1.64.
cc: @GNUDimarik, @JonathanLennox.
[...]

PR #564, opened 2019-01-23
Bouncy Castle 1.67 instead of 1.54 (#562)

Neustradamus:
Bouncy Castle to 1.67 #562 instead of 1.54 with several CVEs...

CVEs: https://www.cvedetails.com/vulnerability-list/vendor_id-7637/Bouncycastle.html


Neustradamus:
@damencho: 1.64 now:
- https://www.bouncycastle.org/latest_releases.html

Neustradamus:
@damencho: 1.65 now:
- https://www.bouncycastle.org/latest_releases.html

Neustradamus:
@damencho: After 1.55/1.56/1.57/1.58/1.59/1.60/1.61/1.62/1.63/1.64/1.65/1.66, there is a 1.67 now:
- https://www.bouncycastle.org/latest_releases.html

damencho:
Have you tested it? Have you done a successful call with jitsi with it? Have you tried it on jigasi, does it work? Does it even compile?

ibauersachs:
It doesn't compile in Jitsi and requires extensive changes in multiple denpendecies, which I've mostly fixed in the new-build branch (and in a draft PR for libjitsi).

PR #592, opened 2019-05-19
Updates smack 4 with latest bundles

damencho:
Updates dependencies and make sure xmpp starts and connects.


damencho:
No idea :) I think there were no significant changes, just bug fixes ... But I haven't tested calling at all in the context of Jitsi Desktop. We use 4.2.4 for jigasi and it works, but jnodes is not used there.

ibauersachs:
Don't bother, we'll find out ;-)

PR #597, opened 2019-06-02
Add PUBLIC_NAT_IP

cbeckr:
A public IP that should be distributed to peers as a contact address
instead of the locally determined one. This is required in the case
of a 1:1 NAT, e.g. within AWS EC2, where the local interfaces do not
correspond with the externally visible address.

Applies to ICE candidates as well as SIP URIs.


cbeckr:
We did a CLA with Atlassian, please let me know if this needs to be renewed with 8x8.

damencho:
Why is that? This had never been a problem when the server does latching for the media.

cbeckr:
This allows us to have Jigasi directly accept incoming calls from Twilio, without requiring any additional servers (FreeSWITCH etc.).
That being said, this may very well be a shortcoming of our configuration or an incompatibility with Twilio.

Issue #613, opened 2019-09-20
Wrong timestamp in RTP when using AMR-WB 


napcode:
We've been using Jitsi for some time with G711 and other codecs. We just tried using AMR-WB but ran into issues. Apparently, Jitsi is sending decreasing RTP timestamps in this configuration. A corresponding PCAP is attached (first few packets of the call including SIP). Rendering of the incoming stream did not work either. Are there any helpful logs for that? PCAP looks okay in this direction.  

Please let me know if more information is needed.
[amr-wb.zip](https://github.com/jitsi/jitsi/files/3636559/amr-wb.zip)

traud:
I am not a team member but an external contributor. Anyway, I am able to confirm your finding with Jitsi Desktop 2.10 and the current nightly of 2.11. The issue is within LibJitsi. Until my change is merged, you have to build LibJitsi and Jitsi Desktop[ yourself…](https://community.jitsi.org/t/18436)

> rendering of the incoming stream did not work

My change just fixes Jitsi as **sender**. That fixed all my issues. Therefore, what means ‘incoming’ in this context: Is your Jitsi Desktop not able to play AMR-WB as **receiver** as well? In that case, which app do you use on the other side?

napcode:
Thanks for looking into it! Sorry I forgot to update this report. We've found an issue on our end as well. It most likely caused the rendering issue I was referring to. I'll try to test both directions using the patch as soon as I can.

traud:
Yes, please, test as soon as possible because I am not a constant contributor to this project and might be off in a few days again. By the way, just out of curiosity: Why don’t you use G.722 instead of AMR-WB?

Issue #540, opened 2018-10-22
Local camera not working locally


weiss:
I'm running Jitsi 2.11.5611 installed from your repository on Debian stretch.  `Options` &rarr; `Video` lists and enables the local camera, but the picture remains black and I see the following message:

```
SEVERE: [96] org.jitsi.impl.neomedia.codec.video.HFlip.log() avfilter_graph_config: -22,
             format AVFRAME, 640x480, class org.jitsi.impl.neomedia.codec.video.AVFrame,
             pixFmt 1, deviceSystemPixFmt 0x56595559
```

The same happens when enabling "Show local video" in a video call.  However, the remote side *does* see my video stream.

This is probably due to some change between 2.11.5602 and 2.11.5604: the former version works for me, the latter doesn't.

Not sure this is an actual bug or just an incompatibility with some Debian stretch component.  The issue described in https://community.jitsi.org/t/camera-not-working-with-jitsi-desktop/15038/ sounds similar, but I'm not sure it's the same thing.

damencho:
2.11.5602 was been openh264 was introduced, so maybe something about hflip as the message says...

traud:
Exactly the same error here on Ubuntu 19.10. For me, also incoming video was affected. LibJitisi tries to flip the picture horizontally, to create a mirror-like user experience in the video settings. LiJitsi flips the picture of the remote party, so right is right and left is left.

Error -22 [seems](http://ffmpeg.org/doxygen/trunk/error_8h_source.html) to be `INPUT_AND_OUTPUT_CHANGED`. Because I have no idea what that means (or why it happens), I applied the patch from within `resources/install/debian/patches/remove-hflip`.

traud:
Mhm. This is not generic to all Debian/Ubuntu because Jitsi Desktop 2.11 works in Ubuntu 16.04 LTS (FFmpeg 2.8). There it still does a HFlip. Is there any documentation where that HFlip source code and/or parameters were coming from? Which is the main platform for Jitsi Desktop (Windows 8.1, Windows 10 Version xxyy, …) and which FFmpeg version is bundled for the other platforms exactly?

damencho:
> Is there any documentation where that HFlip source code and/or parameters were coming from?

That is in libjitsi. https://github.com/jitsi/libjitsi/blob/master/src/org/jitsi/impl/neomedia/codec/video/HFlip.java#L26



[...]

Issue #400, opened 2017-09-12
jitsi fails to find libjawt.so on ubuntu linux using the provided packages 


callegar:
Launching jitsi on debian/ubuntu results in an error being reported on stderr:

```
2017-09-12 10:05:47.066 SEVERE: [21] org.jitsi.impl.neomedia.device.DeviceConfiguration.log() Failed to register custom Renderer org.jitsi.impl.neomedia.jmfext.media.renderer.video.JAWTRenderer with JMF.
java.lang.UnsatisfiedLinkError: /tmp/jna--172125263/jna3728938959406524053.tmp: libjawt.so: cannot open shared object file: No such file or directory
```

This seems to be caused by the fact that the jitsi startup shell script `/usr/bin/jitsi` changes the LD_LIBRARY_PATH and forgets to re-add the jre library path (that is `usr/lib/jvm/java-8-openjdk-amd64/jre/lib/amd64/` in my system) to it.

Editing the script to assure that the jre library path is included in LD_LIBRARY_PATH prevents the "SEVERE" error from being issued at startup.

I have experimented by editing `/usr/bin/jitsi` to add `usr/lib/jvm/java-8-openjdk-amd64/jre/lib/amd64/` to LD_LIBRARY_PATH and this works. For the sake of rapidity, I have used a hardwired path, but the script should probably look at which jre is being used and where its lib dir is in the filesystem.

ibauersachs:
This can't be a generic problem:
- The script extends LD_LIBRARY_PATH (line 39), it doesn't replace it
- It works fine on e.g. Ubuntu 16.04

We'll also need your entire [logs](https://jitsi.org/logs). Or, at the very least:
- the version of Jitsi you were using
- the property value of
  - sun.boot.library.path
  - java.library.path
  - jna.library.path

callegar:
The problem seems to be rather common. Searching jitsi and libjawt on google returns many results. Interestingly, if you look at bug reports for many distros, even when reporting other problems, if the stderr output is included you see that there was a failure in finding libjawt.so. For instance:

http://users.sip-communicator.dev.java.narkive.com/1In2eKM6/jitsi-users-jitsi-can-draw-no-graphics-on-ubuntu-12-04
https://github.com/jitsi/jitsi-meet/issues/242
http://lists.jitsi.org/pipermail/users/2014-February/006365.html

Incidentally, I am on Kubuntu 17.04.

[...]

PR #620, opened 2019-11-11
Updates debian dependencies.

traud:
required for Debian 10, Ubuntu 18.10, Ubuntu 19.04, and Ubuntu 19.10 already

### DIFF
--- a/resources/install/debian/control.tmpl
+++ b/resources/install/debian/control.tmpl
@@ -59,16 +59,16 @@ Standards-Version: 3.9.5
 Package: _PACKAGE_NAME_
 Architecture: i386 amd64
 Depends: default-jre | java7-runtime,
- libavformat58 | libavformat57 | libavformat-ffmpeg56,
- libavcodec58 | libavcodec-extra58 | libavcodec57 | libavcodec-extra57 | libavcodec-ffmpeg56 | libavcodec-ffmpeg-extra56,
- libavfilter7 | libavfilter-extra7 | libavfilter6 | libavfilter-extra6 | libavfilter-ffmpeg5,
- libavutil56 | libavutil55 | libavutil-ffmpeg54,
- libswscale5 | libswscale4 | libswscale-ffmpeg3,
  libappindicator1,
  libxss1,
  libspeex1,
  libspeexdsp1,
  jitsi-archive-keyring
+Recommends:  libavformat58 | libavformat57 | libavformat-ffmpeg56,
+ libavcodec58 | libavcodec-extra58 | libavcodec57 | libavcodec-extra57 | libavcodec-ffmpeg56 | libavcodec-ffmpeg-extra56,
+ libavfilter7 | libavfilter-extra7 | libavfilter6 | libavfilter-extra6 | libavfilter-ffmpeg5,
+ libavutil56 | libavutil55 | libavutil-ffmpeg54,
+ libswscale5 | libswscale4 | libswscale-ffmpeg3
 Description: VoIP and Instant Messaging client
  _APP_NAME_ is an application that allows you to do audio/video
  conversations over the Internet through the SIP and XMPP/Jabber protocol,
--- a/resources/install/debian/control.tmpl
+++ b/resources/install/debian/control.tmpl
@@ -59,11 +59,11 @@ Standards-Version: 3.9.5
 Package: _PACKAGE_NAME_
 Architecture: i386 amd64
 Depends: default-jre | java7-runtime,
- libavformat57 | libavformat-ffmpeg56,
- libavcodec57 | libavcodec-extra57 | libavcodec-ffmpeg56 | libavcodec-ffmpeg-extra56,
- libavfilter6 | libavfilter-extra6 | libavfilter-ffmpeg5,
- libavutil55 | libavutil-ffmpeg54,
- libswscale4 | libswscale-ffmpeg3,
+ libavformat58 | libavformat57 | libavformat-ffmpeg56,
+ libavcodec58 | libavcodec-extra58 | libavcodec57 | libavcodec-extra57 | libavcodec-ffmpeg56 | libavcodec-ffmpeg-extra56,
+ libavfilter7 | libavfilter-extra7 | libavfilter6 | libavfilter-extra6 | libavfilter-ffmpeg5,
+ libavutil56 | libavutil55 | libavutil-ffmpeg54,
+ libswscale5 | libswscale4 | libswscale-ffmpeg3,
  libappindicator1,
  libxss1,
  libspeex1,
--- a/resources/install/debian/control.tmpl
+++ b/resources/install/debian/control.tmpl
@@ -61,14 +61,14 @@ Architecture: i386 amd64
 Depends: openjdk-8-jre-headless | java8-runtime-headless | default-jre-headless,
  libxss1,
  libspeex1,
- libspeexdsp1,
- jitsi-archive-keyring
+ libspeexdsp1
 Recommends: libappindicator1,
  libavformat57 | libavformat-ffmpeg56,
  libavcodec-extra57 | libavcodec57 | libavcodec-ffmpeg-extra56 | libavcodec-ffmpeg56,
  libavfilter6 | libavfilter-extra6 | libavfilter-ffmpeg5,
  libavutil55 | libavutil-ffmpeg54,
- libswscale4 | libswscale-ffmpeg3
+ libswscale4 | libswscale-ffmpeg3,
+ jitsi-archive-keyring
 Description: VoIP and Instant Messaging client
  _APP_NAME_ is an application that allows you to do audio/video
  conversations over the Internet through the SIP and XMPP/Jabber protocol,
--- a/resources/install/debian/control.tmpl
+++ b/resources/install/debian/control.tmpl
@@ -6,7 +6,7 @@ Uploaders: Emil Ivov <emcho@jitsi.org>
 Homepage: _APP_WEB_
 Build-Depends: debhelper (>= 9), javahelper,
                dpkg-dev (>= 1.16.1),
-               openjdk-8-jdk | java8-jdk | java8-sdk | default-jdk,
+               openjdk-8-jdk | java8-sdk,
                ant,
                ant-optional,
                ant-contrib-cpptasks,
@@ -58,7 +58,7 @@ Standards-Version: 3.9.5
 
 Package: _PACKAGE_NAME_
 Architecture: i386 amd64
-Depends: openjdk-8-jre-headless | java8-runtime-headless | default-jre-headless,
+Depends: openjdk-8-jre-headless | java8-runtime-headless,
  libxss1,
  libspeex1,
  libspeexdsp1
--- a/resources/install/debian/control.tmpl
+++ b/resources/install/debian/control.tmpl
@@ -6,7 +6,7 @@ Uploaders: Emil Ivov <emcho@jitsi.org>
 Homepage: _APP_WEB_
 Build-Depends: debhelper (>= 9), javahelper,
                dpkg-dev (>= 1.16.1),
-               default-jdk | java7-sdk,
+               openjdk-8-jdk | java8-jdk | java8-sdk | default-jdk,
                ant,
                ant-optional,
                ant-contrib-cpptasks,
@@ -58,17 +58,17 @@ Standards-Version: 3.9.5
 
 Package: _PACKAGE_NAME_
 Architecture: i386 amd64
-Depends: default-jre | java7-runtime,
- libappindicator1,
+Depends: openjdk-8-jre-headless | java8-runtime-headless | default-jre-headless,
  libxss1,
  libspeex1,
  libspeexdsp1,
  jitsi-archive-keyring
-Recommends:  libavformat58 | libavformat57 | libavformat-ffmpeg56,
- libavcodec58 | libavcodec-extra58 | libavcodec57 | libavcodec-extra57 | libavcodec-ffmpeg56 | libavcodec-ffmpeg-extra56,
- libavfilter7 | libavfilter-extra7 | libavfilter6 | libavfilter-extra6 | libavfilter-ffmpeg5,
- libavutil56 | libavutil55 | libavutil-ffmpeg54,
- libswscale5 | libswscale4 | libswscale-ffmpeg3
+Recommends: libappindicator1,
+ libavformat57 | libavformat-ffmpeg56,
+ libavcodec-extra57 | libavcodec57 | libavcodec-ffmpeg-extra56 | libavcodec-ffmpeg56,
+ libavfilter6 | libavfilter-extra6 | libavfilter-ffmpeg5,
+ libavutil55 | libavutil-ffmpeg54,
+ libswscale4 | libswscale-ffmpeg3
 Description: VoIP and Instant Messaging client
  _APP_NAME_ is an application that allows you to do audio/video
  conversations over the Internet through the SIP and XMPP/Jabber protocol,


damencho:
You know that this will not solve the problem that you will not be able to load ffmpeg there, right?
This will only solve the problem with installation, but media will not work.

traud:
No, because ‘everything’ worked for me. Which media is affected exactly?

damencho:
Video calls? Do you say that the binaries which are linked to libavcodec57 work fine with libavcodec58? And everything is resolved and working. 
What is the output of ldd libjnffmpeg.so, should be somewhere in /usr/share/jitsi/ ....

traud:
Where do I get that library? I do not have it, when I do a clean install. I do get it in ~/.jitsi/sip-communicator.bin when I install the latest stable and then the latest nightly (5618). However, even in Ubuntu 18.04 LTS or when I re-download it from [here](https://github.com/jitsi/jitsi/tree/5716bdae1b723675195eed99b995ec84cdea7fd0/lib/native/linux-64), that library just points to the standard C libs.

damencho:
This is the old static linked library. Oh the lib is inside jar file you will not find it, my fault. You can download it from https://github.com/jitsi/jitsi-lgpl-dependencies/tree/master/src/main/resources/linux-x86-64

damencho:
On that link there are two libs one linked to older ffmpeg and one to the newer, the versions you see in the control file from master.

traud:
Mhm, now I see. `ldd` gives ‘not found’. I have to dig deeper in all this the next week anyway, because I have an issue with the audio codec AMR-WB. Since the Nightlies, AMR-WB does not work at all, even in Ubuntu 18.04 LTS:
```
Cannot build a flow graph with the customized options:
  Unable to transcode format: LINEAR, 44100.0 Hz, 16-bit, Mono, LittleEndian, Signed
    to: AMR-WB/rtp, 16000.0 Hz, Mono
```
That worked in the latest stable (although with many issues as well).

damencho:
AMR was removed from the latest binaries. 
And to make new ffmpeg binaries that they work for new ffmpeg we need to recompile and add them ... not sure when this will happen.

damencho:
The AMR was disabled in this PR https://github.com/jitsi/libjitsi/pull/403 
https://github.com/jitsi/libjitsi/pull/403/commits/fe0f44aa8a36023c1b69edd8122844608f403b27#diff-8b65e7414388ba981937ebdcdc2dc8e1R111
So if you need it, you will need to add it to ffmpeg and enable the encoder and decoder classes in libjitsi.

traud:
Puh. That was valuable information! Thanks to those starting points, I am now able to use AMR-WB in Ubuntu 19.10 (FFmpeg 4.x) with the latest Jitsi Desktop. Finally, I can look at traud/asterisk-amr/issues/17 and #613.

I love to help, however, I am unsure how to proceed here. Because Java allows dynamic loading of libraries, and because Jitsi Desktop even works without FFmpeg, I wonder if those dependencies ‘matter’ at all. Matter means here, being required. Would it not make more sense to move those dependencies from Depends to [Recommends](https://www.debian.org/doc/manuals/debian-reference/ch02.en.html#_package_dependencies)? In that case, we could proceed with this Pull Request, leaving the other puzzle pieces untouched, yet.

Beside that, when a media codec like AMR-WB is disabled in libjitsi (or not available because FFmpeg was not linked), it has not to show up in the user interface of Jitsi Desktop (or at least being a disabled/grey option with a tool tip explaining the situation). Otherwise users (like me at the start) get no media, if they enabled AMR-WB for example. I guess, something like this—grey media codecs, when not available—is not broken but not implemented, yet.

ibauersachs:
Although I won't have time/motivation to work on it, I still think a lot about these issues. One issue is that we currently have one Debian repo with binaries, only differentiating for stable/nightly builds. This was fine while we provided all native stuff as statically linked binaries. Going forward, IMO we should reconsider how the repo is set up and the build process in general. Now with Github Actions and Docker, there are new options. E.g.
- Build and publish jitsi-ffmpeg for Windows/Mac and a statically linked Linux version as Maven artifacts. For Linux, build the native part as an additional library and publish it for various distros (debian/stretch, debian/buster, ubuntu/xenial, etc., maybe some RPMs)
- Build Jitsi Desktop with a proper build system (Gradle/Maven, CMake) *for each* target, and publish to the repo corresponding to which it was built with.
- Or just pull all (native) dependencies as source and build within Jitsi Desktop's build system. _Maybe_ use git submodules for that. But still build from scratch for every target.

In short, one binary for all Linux just isn't going to work (anymore).

traud:
Puh. I do not understand a word (what is my role, what should I do, why). Actually, I understand so less, I cannot even ask a clever question. Here, we are about the package for Debian and its derivates. Let us start with this:
> one binary for all [Debian derivates] just isn't going to work

Can you elaborate why that does not work?

Guessing around: Are you unhappy with that LGPL.jar? Currently, it has two shared libraries. It needs a third library. Actually building this manually is terrible easy after figuring out all [puzzle pieces…](https://community.jitsi.org/t/18436) would be nice if somebody could comment, if I made everything right. Especially https://github.com/jitsi/libjitsi/commit/204d1f80913d6baadb572e91a39a75399c19bf43 is causing me trouble—and I do not understand how to use Maven within Jitsi Desktop (to avoid that `cp` and grab any locally installed packages instead).

ibauersachs:
>Let us start with this:
>> one binary for all [Debian derivates] just isn't going to work
> Can you elaborate why that does not work?

Because of the different ffmpeg versions in Ubuntu 16.04 / 18.04. For some time, Debian and Ubuntu used the ffmpeg fork libav, then switched back to ffmpeg proper, but exported the symbols differently. 

The the newer the OS release, the newer ffmpeg. And they deprecated quite some APIs we use(d), and the newer ones aren't yet available in older distros. We could argue to drop support for 16.04, but that doesn't solve all issues.

traud:
Mhm, the current solution with two libs works quite well for FFmpeg 2 and FFmpeg 3. It ‘just’ needs another lib for FFmpeg 4. I tested Ubuntu 16.04 LTS, 18.04 LTS and 19.10 in the last few days and was quite happy. Playing around with different build systems, docker or so, does not make the source code more stable, which has to support all three FFmpeg (or drop one or two).

All I faced, was one issue with `jawtrenderer` which did not resolve `libjawt.so`. I had to add the RPath for Java 8 in the `libjitsi/src/native/build.xml`. Not sure, if I did something wrong because that was an issue on all my three Ubuntu versions.

Nevertheless, beside FFmpeg, the same applies to OpenSSL. Only Ubuntu 16.04 LTS works with the current OpenSSL 1.0.x adapter. Since 18.04 LTS, OpenSSL 1.1.x is used. I was able to patch that as well. However, the resulting package does not have that several-libs-trick like FFmpeg has already. Therefore, the package works only with Ubuntu 16.04 LTS _or_ Ubuntu 18.04 LTS (and newer). Not sure if that OpenSSL is used with Jitsi Desktop at all. Well, because of the dynamic loading/binding in Java, this does not matter either. If OpenSSL is found, it is used. If not, Jitsi Desktop starts anyway and falls back to Bouncy Castle.

Consequently, I am bit confused. You _have already_ a solution to support multiple versions of an external shared library. The question is not the build system (or the resulting artifacts) but whether you want to maintain the source code for several versions. And actually, the changes for OpenSSL 1.1 and FFmpeg 4 were slim (two symbols each); at least when it comes to compile time. I have not tested run time.

ibauersachs:
I haven't properly looked at your patch in the -lgpl repo, but it seemed like you simply disabled something. This won't work at runtime.
The solution with multiple binaries we have now were manually compiled and put together. This was a time consuming task, and since it was probably more than a year ago, neither me nor Damencho would be able to do it again. Updating a minor version, accepting a pull request, etc. are all way to tedious now.

And yes, the same problems apply to the OpenSSL stuff. But that's not used in Jitsi Desktop.

traud:
> This won't work at runtime.

Please, double-check if anything uses those two symbols at all. I tested just AMR-WB and that works. I researched and mentioned the originating relevant commits, so you see what the alternatives are. As I do not use video – I do not even have a camera –, I cannot help further on that.

Anyway, changes – coping with the _latest_ version of an external library – cannot be fixed with another build system. This is simply work. You are about the decision whether you want to support _multiple_ versions of external libraries (or drop FFmpeg all together).

This change here is just about getting Jitsi Desktop installed at all. If the optional parts are moved from `Depends` to `Recommends`, this gets even more forward compatible.

ibauersachs:
> > This won't work at runtime.
> 
> Please, double-check if anything uses those two symbols at all. I tested just AMR-WB and that works. I researched and mentioned the originating relevant commits, so you see what the alternatives are. As I do not use video – I do not even have a camera –, I cannot help further on that.

I'm not sure when I find time for that.

> Anyway, changes – coping with the _latest_ version of an external library – cannot be fixed with another build system. This is simply work. You are about the decision whether you want to support _multiple_ versions of external libraries (or drop FFmpeg all together).

Correct, incorporating the latest changes is just work. But making these changes _available_ without too much manual intervention requires a sane build system. Or rather just _a_ build system. Currently we have neither.

> This change here is just about getting Jitsi Desktop installed at all. If the optional parts are moved from `Depends` to `Recommends`, this gets even more forward compatible.

I'm fine with that.

Actually, given all the issues with the video and screensharing support, it might be an option to completely remove that. Jitsi Meet is the better alternative for that.

traud:
> This won’t work at runtime.

In Linux, late binding happens. Only when the symbol is used, a segmentation fault happens. However, the shared library must be able to resolve its depended libraries = files on disk, path in scope.

I rechecked. None of the two symbols are ever called within the Jitsi universe. At least, I did not find anything via a GitHub search. Of course, an external library might use it. Not sure if you have any ‘contracts’ for that library. If yes, there has to be some versioning like a SONAME bump. Probably it is better to remove those two symbols completely. Then, if there is an issue, it reveals more obvious because all platforms are affected.

> making these changes available without too much manual intervention requires a sane build system

OK. I researched that again and bought myself a camera. Here are my findings: Currently, FFmpeg is used for the media format H.264 _or_ AMR-WB. Some cameras come with native H.264 output. That is not used. FFmpeg comes with its own H.264 implementation. That is not used. Instead, OpenH264 is used through FFmpeg. Therefore, a custom **bundled** FFmpeg is used. That bundled FFmpeg does not come with AMR-WB. Therefore it is H.264 _or_ AMR-WB right now. According to my tests and review of the source code, only those two media formats require FFmpeg. By now (see jitsi/libjitsi#498), these two media formats are detected at runtime. Consequently, this is a no-brainer.

However, FFmpeg is also used to enable camera/video support in general. This is an optional feature. Jitsi Desktop starts and can be used without video, just for audio – at least according to my tests. There is only a cosmetic issue on the command line, where the exception about the unsatisfied link is printed in raw, rather than with a human-readable error message. Consequently, it is OK to move FFmpeg from `Depends` to `Recommends`. I changed my Pull Request accordingly.

Now comes the FFmpeg version issue. Currently, the LGPL shared library bridges FFmpeg 2 and FFmpeg 3. This is done through one shared library for each version for each platform. That sounds a lot. However, this affects just Linux, because all _other_ platforms get that bundled FFmpeg (with OpenH264). In Linux, there are currently two platforms: `i386` and `amd64`. Consequently, those are three versions on two platforms, resulting in six shared libraries. Each library can be built with one single command: `ant ffmpeg -DskipOpenh264=true`.

When do you build those libraries:
A) When there is a bug in the wrapper (happened last ? in Dec. 2017). Or
B) When there is a new version of FFmpeg (happened last ? in Apr. 2018)
The last case happens right now: Besides FFmpeg 2 and FFmpeg 3, FFmpeg 4 needs to be added (which offers H.265 and AV1 support, by the way). Because in the mid-term, the bundled FFmpeg must be updated to the latest version, supporting that new version is not adding any additional overhead. That was clarified above already. The problem is how to keep support for the old versions. This is not a problem, because those old wrapper libraries need to be re-build just in case A. In case B, new ones are added, the old ones are not touched.

I hope this explains why I have problems to understand why an _automatic_ build system is needed. I can offer to build those two wrapper libraries for FFmpeg 4. Actually, for `amd64`, I have it already here. It is just creating a virtual machine, install Debian Buster with the right platform†, issue the above command. Then, those files must be added to the JAR: `mvn package`. Then, this third variant must be loaded dynamically, as well. Done.

† I would not even use those cross-compiling features to build `i386`.

traud:
Now, there are several alternatives to get this Pull Request merged:
(A) move the FFmpeg dependencies from `Depends` to `Recommends`,
(B) beside FFmpeg 2.8 and 3.x, add FFmpeg 4.x to the dependencies,
(C) wait until the FFmpeg 4.x wrapper was added with jitsi/jitsi-lgpl-dependencies#10.

Currently, it seems to me, we wait for all three conditions to fulfill. What about going just for alternative A, first? That would allow the installation for those on newer Linux platforms. They are able to run Jitsi Desktop because video is just optional. And still, the dep file would indicate that FFmpeg 4.x is not supported out-of-the-box, yet.

Then, when condition C is fulfilled, I create a new Pull Request here which requests B. What about that? With that approach we do not have so long lived Pull Requests no one benefits from. Furthermore, it would move the discussion to the repositories where it belongs to.

Actually, there is an alternative D:
In Debian, meta packages exist `libavcodec-extra` and `libavfilter-extra`. Those install all other packages (avutil, avformat, …). I am not a fan of that because† it does not indicate that LibJitsi is not working with those. Anyway, I thought I should mention it. On the other hand, those meta packages would avoid to touch this file anytime soon.
† another reason against D: `libavfilter-extra` does not work for FFmpeg 2.8. One approach for those platforms would be not to go for the lib packages but just for the app package `ffmpeg`. That package installs all FFmpeg packages for Jitsi Desktop for sure. Personally, I do not like that either, either.

traud:
OK, I went for alternative A to get this Pull Request done. On Debian 10 and Ubuntu 20.04 LTS, this change allows Jitsi Desktop to be run, although without video (and AMR-WB) support, yet. With this change, an user is able to simply double-click the `.deb` file after installing [jitsi-archive-keyring](https://download.jitsi.org/unstable/jitsi-archive-keyring_1.0.1_all.deb).

By the way, for what feature is that keyring package required? If we get that optional as well, users are able to install Jitsi Desktop even without Int**er**net. I know companies who use Jitsi Desktop 2.10 in their Int**ra**net.

damencho:
> By the way, for what feature is that keyring package required?

This is just adding the public key for the repo and adding the stable repo to `etc/apt/sources.list.d`. This one can be optional, we used it at some point in the past when changing the Debian repository link so we don't break updates for the old version while transitioning to the new repo.

traud:
Does anything block this request? Recommended commit message:

Debian: Depend on FFmpeg and AppIndicator optionally.
- AppIndicator is enabled optionally via Menu » Tools » Options » General » Tray implementation.
- FFmpeg is used for video and the audio codec AMR-WB (therefore, libavcodec-extra preferred).
- jitsi-archive-keyring was used so updates do not break while transitioning to a newer Debian repository.

Jitsi Desktop runs without those. Furthermore, Jitsi changed to Java 8 as minimum, which is reflected here now.

Banellica:
Dear guys

I need to install Jitsi Client on Ubuntu 18.04 (and it complains: "unmet dependencies: jitsi-archive-keyring not installable").

I'd like to just use the audio chat functionality (no need for video).

I am not an expert. Can you provide a simple prescription - what and how should I install?

traud:
> Ubuntu 18.04 LTS complains: "unmet dependencies: jitsi-archive-keyring not installable"

There, [you go …](http://github.com/jitsi/jitsi/issues/540#issuecomment-600524696)
Alternatively, for Ubuntu 18.04 LTS, install just the [latest keyring](https://download.jitsi.org/stable/jitsi-archive-keyring_1.0.1_all.deb) and then [go for this …](http://github.com/jitsi/jitsi/issues/400#issuecomment-565431096)

PR #628, opened 2019-12-17
Resources (Branding): Avoids repeating Info messages in the log.

traud:
This change avoids the Info message `Missing resource for key: service.gui.MAIN_WINDOW_BACKGROUND` which is printed in the log file, every time when the user moves the cursor over one of the Dial panels.

### DIFF
--- a/src/net/java/sip/communicator/impl/resources/ResourceManagementServiceImpl.java
+++ b/src/net/java/sip/communicator/impl/resources/ResourceManagementServiceImpl.java
@@ -231,8 +231,6 @@ public class ResourceManagementServiceImpl
 
         if (path == null || path.length() == 0)
         {
-            if (logger.isInfoEnabled())
-                logger.info("Missing resource for key: " + urlKey);
             return null;
         }
         return getImageURLForPath(path);


traud:
This is one of those issues, an external contributor can hardly fix:

When you open the dial panel in the main view (that view with the buttons for 1-9, 0, *, and #), or when you open the dial panel while in a call, the log is spammed with the above **Info** message. For _every_ mouse movement, an _additional_ log entry is created.

The problem with this issue: What is the right fix?

The issue was introduced when [image.properties](https://github.com/jitsi/jitsi/commits/master/src/net/java/sip/communicator/impl/gui/utils/images.properties) got removed in April 2008 and that misleading message was reported in [July 2008](https://community.jitsi.org/t/1484) the first time. Actually, all messages related to resources are only for somebody who brands Jitsi Desktop. Those message are just to double-check he added all images, colors, and setting he liked. Right? This is nothing for end-users [not even](https://community.jitsi.org/t/5007/2) for developers. This is just for the branding guy to double-check he has no typing error in the file path.

Right? Consequently, I wonder if not all log messages in that class should be removed, because it is just for the branding guy. Or should it be lowered below Info, then only developers would see them. **Info** is user level. **Warn** is user-_reporting_ level actually.

The above Pull Request is some sort of a minimum because a log entry for each mouse movement is too much, for sure. Again, this sore of issues are terrible difficult to fix for an external contributor because the contributor would have to decide something he cannot decide actually (after the first step: understand the issue, which is difficult enough). This is something, a team member has to listen carefully and re-think the made decision, when something like this comes up in the mailing list, discussion board, or as issue report.

Take the above as proposal to start a discussion to find a decision. Personally, I would remove all logging in that class as I am developer, at maximum.

damencho:
Done in https://github.com/jitsi/jitsi/commit/fb19f667e812ca63c45e9d80c271c011f291f462
Thanks

PR #629, opened 2019-12-22
AppIndicator: Update OSGi after Re-architecting.

traud:
AppIndicator uses org/jitsi/utils/logging/Logger. Therefore, when enabled via Menu » Tools » Options » General » Tray implementation, a NoClassDefFoundError was thrown at the start of Jitsi Desktop.


traud:
All these issues were caused by commit 4cafb3c. I found the issue with AppIndicator while testing, okay. However, I found the next two issues via double-checking thanks to a simple
`grep --recursive 'org.jitsi.utils.logging' .`

The original commit forgot to update **any** manifest file. There have been so many commits since then which (partially) fix that failure of the original commit. I wonder why, after eight months, I still find three broken bundles.

damencho:
> The original commit forgot to update **any** manifest file. There have been so many commits since then which (partially) fix that failure of the original commit. I wonder why, after eight months, I still find three broken bundles.

Cause nobody fully tested all the plugins that were failing. Some of them we use in jigasi, but there the environment does not have this classloader restriction coming from apache Felix.

traud:
> nobody fully tested all the plugins that were failing

Not sure, you got the complain. Commit 4cafb3c was repaired not by one but by _several_ commits. On the very first repair attempt, the correct behavior would have been to undo the commit, inform the original author, wait for someone to update all manifest files, and then re-apply that commit. Even if that offending commit could not been undoed, somebody could have gone through it, and fix all manifests. Now, we have _one_ commit which got repaired by _several_ other commits – and eight months later an external contributor is creating another commit.

The current situation is very complicated for an external contributor like me, because first of all I had to understand the cause. Such a contributor has to learn or know the basics of Apache Felix. And the cause was not the broken manifest but that incomplete commit. The next step would have been to go through that commit and fix all related manifests. However, because the manifest fixes got spread over several (!!!) other commits over months (!!!), I would have to reverse engineer also which manifests got missed by _those_ commits.

I went another approach and went through the whole code statically. I did not test anything either (except the AppIndicator where I started to find that issue). This cause could have been solved **statically** when the first repair-commit was done, no testing was/is required.

Furthermore, I am not sure if more imports are broken. I wonder if anyone of the core members has a tool or script which is able to create those manifests automatically, or at least semi-automatically. I am currently tracing another bug about the Apache Felix classloader – and have no idea how to fix it and what the cause is. The issue seems to be there since day one.

Once again, this is not a lack/failure of testing but of how the Jitsi team approached the offending commit.

damencho:
I know why the initial commit was done. And my answer will be the same as the previous commit. We would try to keep this project working and not break it, but we do not have the resources to test everything and it is normal to miss stuff.
I had gone and had tested and fixed some stuff we missed, but I haven't seen all the problem bundles.
If we had seen problems, we would not stop progress on other projects with reverting that commit. We would fix the problems when time permits and continue.

Thank you for tracking those and fixing them.

Those manifests had always been created by hand. 

What is the other Felix bug, can you open an issue and describe it there?

PR #636, opened 2020-01-16
Refactor reconnect plugin.

damencho:
Now we have a registration state listener per provider in the wrapper.
If there is some locking or slowness it will not affect all providers.
Minimizes the synchronized blocks and make them per provider.

### DIFF
--- a/src/net/java/sip/communicator/plugin/reconnectplugin/PPReconnectWrapper.java
+++ b/src/net/java/sip/communicator/plugin/reconnectplugin/PPReconnectWrapper.java
@@ -1,0 +1,361 @@
+/*
+ * Copyright @ 2018 - present 8x8, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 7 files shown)


damencho:
I've updated the PR with the things we discussed offline.

src/net/java/sip/communicator/plugin/reconnectplugin/PPReconnectWrapper.java
```diff
@@ -358,4 +373,93 @@ public String toString()
...
+
+        /**
+         * Reconnects the provider.
+         */
+        @Override
+        public void run()
+        {
+            if(thread == null || !Thread.currentThread().equals(thread))
+            {
+                thread = new Thread(this);
```
paweldomas: I'm thinking that new thread is not needed for the reconnect task. This thread should be dead already by the time a new reconnect is needed and I see `Timer` instance serving the purpose of this thread. What do you think?

damencho: O you are right since we moved to the per-connection timer. Updating it now.

PR #638, opened 2020-01-21
Apache Felix (OSGi): Allow X11 to the Java Classloader.

traud:
Otherwise, sometimes (not reproducible), JAWT Rendering is not able to load the class sun.awt.x11.XErrorHandlerUtil. Requires the latest LibJitsi, see jitsi/libjitsi#514.


traud:
I think, however I am neither tested not am I sure, `felix.unit.test.properties` should be updated as well. I cannot test `ant test` because it gives me
```
BUILD FAILED
build.xml:499: ./lib/accounts.properties not found - did you copy the template?
```

PR #606, opened 2019-08-21
Replaced org.jitsi.util.StringUtils to org.apache.commons.lang3.StringUtils

GNUDimarik:

PR #655, opened 2020-03-22
Add commons-lang3 to Maven modules, check build with GH Actions

ibauersachs:
Missed in #606 and follow-up to d9d575df6acf638d1ed749538085397a32d3db1e


ibauersachs:
@damencho ping

PR #666, opened 2020-04-03
Adjust jitsi.spec to build RPM packages again.

ckujau:
* Removed commented out macro to prevent:
  warning: Macro expanded in comment on line 133: %{_libdir}/jitsi/

* Removed ant-nodeps build dependency to prevent:
  error: Failed build dependencies:
          ant-nodeps is needed by jitsi-2.9-5537.x86_64


mathieujobin:
should the version be automatic ? it seems we are at 2.14.x 

and weirdly, 2.14.721 was released months before 2.14.695

PR #687, opened 2020-05-21
fix: Fixes hanging up incoming calls with media.

damencho:


damencho:
Adds:
```
            || CallPeerState.CONNECTING_INCOMING_CALL.equals(peerState)
            || CallPeerState.CONNECTING_INCOMING_CALL_WITH_MEDIA.equals(peerState)
```
to the cases of hanging up call

damencho:
And adding one more hangup cause internal server error.

PR #700, opened 2020-07-16
Add room destroyed event for local user

theunafraid:
This PR adds the LOCAL_USER_ROOM_DESTROYED event to notify the current user that the joined room is no more.


src/net/java/sip/communicator/impl/protocol/jabber/ChatRoomJabberImpl.java
```diff
@@ -227,7 +227,9 @@ public ChatRoomJabberImpl(MultiUserChat multiUserChat,
             new SmackSubjectUpdatedListener());
         multiUserChat.addMessageListener(new SmackMessageListener());
         multiUserChat.addParticipantStatusListener(new MemberListener());
-        multiUserChat.addUserStatusListener(new UserListener());
+
+        UserListener userListener = new UserListener(this);
+        multiUserChat.addUserStatusListener(userListener);
```
damencho, line 232: This change is not needed, right?

theunafraid:
Not needed will be using LOCAL_USER_LEFT.

PR #755, opened 2021-12-20
fix: Fix checking for lobby error.

damencho:
Leaves the fallback to old code, till everything is updated. We need to drop it at some point.

### DIFF
--- a/src/net/java/sip/communicator/impl/protocol/jabber/ChatRoomJabberImpl.java
+++ b/src/net/java/sip/communicator/impl/protocol/jabber/ChatRoomJabberImpl.java
@@ -255,7 +255,13 @@ public class ChatRoomJabberImpl
         {
             if (packet != null)
             {
-                ExtensionElement lobbyExtension = packet.getExtension("lobbyroom", "jabber:client");
+                ExtensionElement lobbyExtension = packet.getError().getExtension("lobbyroom", "jabber:client");
+
+                // let's fallback to old code if this is missing, TODO: drop this at some point
+                if (lobbyExtension == null)
+                {
+                    lobbyExtension = packet.getExtension("lobbyroom", "jabber:client");
+                }
 
                 if (lobbyExtension instanceof StandardExtensionElement)
                 {
--- a/src/net/java/sip/communicator/impl/protocol/jabber/ChatRoomJabberImpl.java
+++ b/src/net/java/sip/communicator/impl/protocol/jabber/ChatRoomJabberImpl.java
@@ -255,7 +255,8 @@ public class ChatRoomJabberImpl
         {
             if (packet != null)
             {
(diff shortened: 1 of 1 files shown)


damencho:
@ibauersachs We talked about it at some point :) I think this is good.
https://github.com/jitsi/jitsi-meet/pull/10673
https://github.com/jitsi/lib-jitsi-meet/pull/1830

For now the error will be: 
```
<presence to='9666aa17-64c4-4035-b85f-3bd3716a804c@damencho.example.com/lMnS0Dw3'
          from='damencho@conference.damencho.example.com/9666aa17' xmlns='jabber:client' type='error'>
    <error type='auth' code='407'>
        <registration-required xmlns='urn:ietf:params:xml:ns:xmpp-stanzas'/>
        <lobbyroom>damencho@lobby.damencho.example.com</lobbyroom>
    </error>
    <lobbyroom>damencho@lobby.damencho.example.com</lobbyroom>
    <x xmlns='http://jabber.org/protocol/muc'/>
</presence>
```
And in the future, we can drop the <lobbyroom> that is directly under error and the code for the fallback in ljm and jitsi.
Is this ok?

ibauersachs:
This works, but I'd prefer to give `lobbyroom` a proper namespace instead of accidentally inheriting `jabber:client`. And it would enable parsing using a properly defined `LobbyRoomExtensionElement` in jitsi-xmpp-extensions.

damencho:
I see, I will check that.

damencho:
Like this, right?

```
<presence type='error' to='151b15e4-1de6-44f4-a430-dfe7010a164e@damencho.example.com/qtuWrC14'
          from='damencho@conference.damencho.example.com/151b15e4' xmlns='jabber:client'>
    <error type='auth' code='407'>
        <registration-required xmlns='urn:ietf:params:xml:ns:xmpp-stanzas'/>
        <lobbyroom xmlns='http://jitsi.org/jitmeet'>damencho@lobby.damencho.example.com</lobbyroom>
    </error>
    <lobbyroom>damencho@lobby.damencho.example.com</lobbyroom>
    <x xmlns='http://jabber.org/protocol/muc'/>
</presence>
```

ibauersachs:
Yes. Do we still need the `x`-element?

damencho:
I suppose so, this is how it is in prosody: 
https://github.com/damencho/prosody/blob/1e42cdef094e3c34d0c3791d5d2b55d1726d8313/plugins/muc/members_only.lib.lua#L126
I haven't gone into details :) ...

ibauersachs (approved):
The code for the new-build branch will look slightly different with the Smack upgrade there.

damencho:
Yep, it is more important to push it in ljm so it will start updating on clients :)

PR #764, opened 2022-05-21
Fix installers and repo information

ibauersachs:


ibauersachs:
@damencho Can you please check if the DMG from https://github.com/jitsi/jitsi/releases/tag/untagged-92026f7422c4200b07b9 works?

damencho:
I tried it on M1 mac and it starts. No audio & video devices in settings. Do you want me to try on the x86 mac?
[jitsi.log](https://github.com/jitsi/jitsi/files/10779410/jitsi.log)

damencho (approved):

ibauersachs:
libjitsi has no Mac/aarch64 build yet, so the missing devices are ok. Thanks!

PR #766, opened 2022-08-12
Updates to latest libjitsi.

damencho:


modules/plugin/update/src/main/java/net/java/sip/communicator/plugin/update/UpdateServiceImpl.java
```diff
@@ -249,21 +249,12 @@ private static boolean isLatestVersion()
                      */
                     if (downloadLink != null)
                     {
-                        if (OSUtils.IS_LINUX32)
-                        {
-                            downloadLink
-                                = downloadLink.replace("amd64", "i386");
-                        }
-                        else if (OSUtils.IS_LINUX64)
+                        if (OSUtils.IS_LINUX)
```
ibauersachs, line 252: This one is a bit fishy, but irrelevant for jigasi. I think the correct changes/checks are in the installers branch.

ibauersachs (approved):

PR #771, opened 2022-09-29
make deb-prepare.sh compatible with Debian

Af0x:
Created a variable to optionally adjust the package names for Ubuntu.


ibauersachs:
Fixed in d1f7cbae0db310ab98b2b4b961c7c2a4de0a14c4

PR #773, opened 2022-10-01
Make the gpg-signing-key a parameter

Af0x:


ibauersachs:
Please update the callers of this script to pass the argument.
This is probably from within the GitHub Actions.

PR #786, opened 2023-04-19
remove task comment

jelgun:
This PR removes a task comment (i.e. a comment referring to a work that could/should be done in the future or was already done). Instead, an issue was created.


ibauersachs:
Thanks for the idea, but please stop. I'd rather have the TODOs instead of issues nobody is looking at.

PR #795, opened 2023-05-04
README: There is no stable repository, just unstable.

traud:
Previously, the README mentioned two repositories, one stable and one unstable. However, the stable one is empty and therefore not used. Furthermore, clarified what the term 'distro-name' refers to.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,11 +30,11 @@ Download the installers
 from [GitHub releases](https://github.com/jitsi/jitsi/releases/latest/).
 
 #### Debian/Ubuntu
-An APT repository is available at https://nexus.ingo.ch/jitsi-desktop/.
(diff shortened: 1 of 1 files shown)


ibauersachs:
Well, currently you're correct. But it's all prepared so that a release can be promoted to the stable repo. Not sure it's worth removing the info now.

ibauersachs:
> By the way, the [stable folder](https://nexus.ingo.ch/jitsi-desktop/) contains a very large, 4 GByte file called `oci`. Not sure if that is intended.

It is, please drop me an e-mail if you want to know more.

traud:
> it's all prepared so that a release can be promoted to the stable repo

That is OK. However, then the example below should be consistent and use `stable` as well. Furthermore, until that release, I would use `unstable` for both. I took me quite a while to understand this. And at the start, I went for the not yet used repository.

PR #798, opened 2023-05-09
Debian: Allow Jitsi to be started in Terminal

traud:
Furthermore, do not require Bash but just any POSIX shell.

### DIFF
--- a/resources/install/debian/jitsi
+++ b/resources/install/debian/jitsi
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/sh
 
 SHOW_SPLASH=true
 SPLASH_ARG=""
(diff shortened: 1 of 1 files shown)


traud:
Actually, the problem was the `echo`; if started from the Terminal, Jitsi Desktop started from background and no System.out was printed to the Terminal. Because I removed the `echo`, the `exec` was not needed anymore. Not sure why `command -v` and `bash` were used/required. I lifted that as well. Tested for [POSIX compliance …](https://unix.stackexchange.com/q/48786)

ibauersachs:
The echo was proabably debugging stuff that I forgot to remove. `command -v` was for some reason necessary, you could probably search old PRs/issues and find something (probably with `which` though).

traud:
That file was created with the Gradle build commit. When looking at its [parent commit](https://github.com/jitsi/jitsi/tree/345a7c83911984d049eeb88f00958cc8ed60c2ea/resources/install), I do not find any shell command file which uses `which` or `command -v`. Are you saying that there was an issue report … and I should keep that `command -v java` so this Pull Request get merged?

ibauersachs:
It was introduced in d86bf7cc8955bdbdd758a8161fdb8abbb62ef7f6. I had some memories of there being a bug report, but it's possible this isn't relevant anymore (or never was).

PR #812, opened 2023-12-13
Update installers.yml from 23.04 to 23.10

mathieujobin:


mathieujobin:
@ibauersachs do you think we should update this to Mantic now that Lunar is deprecated?

I am looking forward testing the latest version, ahead of the next LTS.

PR #825, opened 2025-04-08
Update github action versions

JonathanLennox:


mathieujobin:
I am so looking forward new builds for the desktop version... I love it as my VoIP of choice... but its been working less and less well on new Linux

PR #844, opened 2026-05-08
Skip deb package signing on PR builds; use rm -f for man-db auto-update removal

JonathanLennox:
Secrets (GPG_KEY, GPG_PW) aren't available on PRs from forks, causing the GPG import step to fail with "no valid OpenPGP data found". Guard the import on master only and skip debsign when no GPG ID is provided so PRs still validate the deb build.

The man-db auto-update file no longer exists on current GitHub runner images, so the unconditional rm fails the build. Use -f so the step is idempotent across runner image versions.

### DIFF
--- a/.github/workflows/installers.yml
+++ b/.github/workflows/installers.yml
@@ -135,11 +135,12 @@ jobs:
       - name: Install tools
         run: |
           # No need for updated man pages on GitHub
-          sudo rm /var/lib/man-db/auto-update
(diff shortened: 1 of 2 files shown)


damencho (approved):

JonathanLennox:
Closing in favor of https://github.com/jitsi/jitsi/pull/845, since CI builds still don't work properly across forks.

